PR #77, opened 2014-08-12
Added windows CI

karthik:
This PR will add Windows CI to the package. So in addition to what Travis accomplishes, you'll get something similar from `appveyor`. You may not want to merge for now, at least until you can fix the issue that make the build fail (it does now). I'll leave it up to the Toby (or anyone else) whether or not to merge this.


cpsievert:
I'm seeing an appveyor failure in #226. Not sure what exactly is causing it, but closing this temporarily to see if it fixes it.

Issue #26, opened 2014-05-12
Bad filename response with directories


noqqe:
When you set a directory within the `filename` like `directory/file` the API returns the directory two times.  `directorydirectory/file`

When you use this as a simple example:

```
library(plotly)

py <- plotly(username="noqqe", key="MYKEY")

x <- rnorm(500)
data = list(x = x,
            type = 'histogramx')
l <- list(autosize = FALSE,
          width = 600,
          height = 400,
          showlegend = FALSE)

response <- py$plotly(data, kwargs = list(layout = l, filename="Mountainbike/TEST"))

filename <- response$filename
print(filename)
```

You recieve the following

```
$ Rscript lolo.R 
Loading required package: RCurl
Loading required package: methods
Loading required package: bitops
Loading required package: RJSONIO
Loading required package: ggplot2
[1] "MountainbikeMountainbike/TEST"
```

mkcor:
Good catch, thanks @noqqe !

PR #156, opened 2015-01-20
Fix response filepath when contains directory

mkcor:
This PR addresses issue #26.

What's posted is actually fine (resulting tree in plotly account is what you expect online).

Value `kwargs$filename` is preserved under `toJSON()` but then under `postForm()` some parsing issue generates the double directory prepending...  Which in turns propagates to `fromJSON()` -- I'm talking about https://github.com/ropensci/plotly/blob/master/R/plotly.R#L107

That's why I suggest overwriting `resp$filename` with the initial `kwargs$filename` so the returned response is what you expect -- no impact on what's actually sent to plotly (as the web platform).

Note that the resulting plotly plot is private by default -- is this what we expect?

/cc @sckott @chriddyp

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: plotly
 Type: Package
 Title: Interactive, publication-quality graphs online.
-Version: 0.5.17
+Version: 0.5.18
 Authors@R: c(person("Chris", "Parmer", role = c("aut", "cre"),
     email = "chris@plot.ly"),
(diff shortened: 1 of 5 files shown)


mkcor:
@theenginear *bambi eyes*

theengineear:
Looks reasonable and it has the tests. :tada:

mkcor:
Thanks!

Updating version number and `NEWS` file...

PR #158, opened 2015-01-22
Support as.Date() conversion within ggplot code

mkcor:
Sorry, the branch name has become unrelated to the PR.

`inherits()` checks the class.  It's pretty frustrating that I cannot just do
```r
is.null(class(myObject))
```
!! Class is returned as a string, so it's `"NULL"`, not `NULL`. Line 52, I'm basically checking for the existence of dimension `x.name` or `y.name`. (Don't be confused, the period does not mean anything here!)

The whitespaces are just for styling.

Thanks! @theengineear

### DIFF
--- a/R/trace_generation.R
+++ b/R/trace_generation.R
@@ -352,14 +352,14 @@ toBasic <- list(
 #' @export
 #' @return list of geom info.
 #' @author Toby Dylan Hocking
-group2NA <- function(g, geom){
+group2NA <- function(g, geom) {
   poly.list <- split(g$data, g$data$group)
   is.group <- names(g$data) == "group"
   poly.na.df <- data.frame()
-  for(i in seq_along(poly.list)){
-    no.group <- poly.list[[i]][,!is.group,drop=FALSE]
-    na.row <- no.group[1,]
-    na.row[,c("x", "y")] <- NA
+  for (i in seq_along(poly.list)) {
+    no.group <- poly.list[[i]][, !is.group, drop=FALSE]
+    na.row <- no.group[1, ]
+    na.row[, c("x", "y")] <- NA
     poly.na.df <- rbind(poly.na.df, no.group, na.row)
   }
   g$data <- poly.na.df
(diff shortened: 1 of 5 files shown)


R/trace_generation.R
```diff
@@ -41,13 +41,24 @@ layer2traces <- function(l, d, misc) {
...
-        data.vec <- l$data[[col.name]]
+        dtemp <- l$data[[col.name]]
+        if (is.null(dtemp)) {
+          if (!inherits(g$data[[paste0(a, ".name")]], "NULL")) {
+            # Handle the case where as.Date() is passed in aes argument.
+            if (class(g$data[[a]]) != class(g$data[[paste0(a, ".name")]])) {
+              g$data[[a]] <- g$data[[paste0(a, ".name")]]
+              data.vec <- g$data[[a]]
+            }
+          }
```
theengineear: There are two if statements without an else. is something bad going to happen if `data.vec` isn't set?

mkcor: Yes, well, if `data.vec` is not set, it's `NULL` and it's still okay because I `try()` in subsequent operations (https://github.com/ropensci/plotly/blob/master/R/trace_generation.R#L56, etc.).  So the exceptions are handled.

My initial intention was to leave line 50
```r
data.vec <- l$data[[col.name]]
```
and follow with
```r
if (is.null(data.vec)) {
  if (!inherits(g$data[[paste0(a, ".name")]], "NULL") && (class(g$data[[a]]) != class(g$data[[paste0(a, ".name")]]))) {
    g$data[[a]] <- g$data[[paste0(a, ".name")]]
    data.vec <- g$data[[a]]
  }
}
```
The `else` being, well, leave it alone (it will remain `NULL`).  Do you think this would convey more meaning?

The two `if`s really mean one condition (i.e., the two columns are of different non-null classes) but I split them for readability, also to save one check if the first condition is not even met.

theengineear:
@mkcor , if the logic is ok with it being left undefined, its seems good : D

:tada:

PR #165, opened 2015-02-09
Remove files related to image tests

mkcor:
Always wanted to keep these outside the R package.

They live on branch `add-r-cookbook-tests`.


tdhock:
+1

PR #167, opened 2015-02-18
Carson scales free

cpsievert:
Howdy @chriddyp @mkcor @tdhock!

The purpose of this pull request is to fully implement `facet_wrap(..., scales = 'free')`.

It also seems possible to implement space = "free", but perhaps that should be in a separate pull request.

PR #172, opened 2015-02-28
Toby cookbook-axes

tdhock:
This branch translates some of Marianne's visual tests from the axes.R file in the add-r-cookbook-tests branch 

https://github.com/ropensci/plotly/blob/add-r-cookbook-tests/tests/cookbook-test-suite/axes.R

to the format I need for adding them to the test table.

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -311,7 +311,12 @@ gg2list <- function(p){
     ax.list$title <- if(length(scale.i)){
       sc <- p$scales$scales[[scale.i]]
       trace.order.list[[xy]] <- sc$limits
-      trace.name.map[sc$breaks] <- sc$labels
+      if(is.character(sc$breaks)){
+        if(is.character(sc$labels)){
+          trace.name.map[sc$breaks] <- sc$labels
+        }
+        ##TODO: if(is.function(sc$labels)){
+      }
       if (is.null(sc$breaks)) {
         ax.list$showticklabels <- FALSE
         ax.list$showgrid <- FALSE
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -274,6 +274,12 @@ gg2list <- function(p){
   trace.name.map <- c()
   for(xy in c("x","y")){
     ax.list <- list()
+    coord.lim <- p$coord$limits[[xy]]
+    if(is.numeric(coord.lim)){
+      ## TODO: maybe test for more exotic coord specification types
+      ## involving NA, Inf, etc?
+      ax.list$range <- coord.lim
+    }
     s <- function(tmp)sprintf(tmp, xy)
     ax.list$tickcolor <- toRGB(theme.pars$axis.ticks$colour)
     
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -160,7 +160,10 @@ gg2list <- function(p){
         names(ranks) <- br
         misc$breaks[[sc$aesthetics]] <- ranks
       }
-      misc$trans[sc$aesthetics] <- sc$trans$name
+      ## store if this is a reverse scale so we can undo that later.
+      if(is.character(sc$trans$name)){
+        misc$trans[sc$aesthetics] <- sc$trans$name
+      }
     }
     reverse.aes <- names(misc$trans)[misc$trans=="reverse"]
     
--- a/tests/testthat/test-cookbook-axes.R
+++ b/tests/testthat/test-cookbook-axes.R
@@ -48,7 +48,8 @@ test_that("scale_y(limits) hides points", {
   
 bp.coord <- bp + coord_cartesian(ylim=c(5, 7.5))
 test_that("Using coord_cartesian zooms into the area", {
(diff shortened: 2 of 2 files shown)


tdhock:
TODO: add tests in R code for things like font color.

tdhock:
Test table for this branch is

http://ropensci.github.io/plotly-test-table/tables/fa8ab30229ca0491580b218f32b551738c7adfd2/index.html

most tests are looking good except

http://ropensci.github.io/plotly-test-table/tables/fa8ab30229ca0491580b218f32b551738c7adfd2/cookbook-axes-coord-ylim.html

chriddyp:
nice! are we planning on adding the rest of the cookbook `.R` files in separate PRs?

tdhock:
well I was thinking of proposing 1 PR per cookbook .R file, and adding fixes in the same PR along the way, so that master is always passing checks on travis.

chriddyp:
OK, cool, that makes sense!

tdhock:
Hey @cpsievert I did git pull origin master from this branch, and I was expecting that travis would run and redo the test table, but it does not seem to be running. Any idea why? If you know how to fix it then go ahead and push to this branch.

cpsievert:
Travis doesn't trigger when you pull or do anything locally. Have you tried `git push origin toby-cookbook`?

cpsievert:
Nevermind, looking into now...

cpsievert:
I think we just have to be patient. Travis is still building...

cpsievert:
It's up now -- http://ropensci.github.io/plotly-test-table/

tdhock:
awesome, but indeed there are some missing images in the current test table

http://ropensci.github.io/plotly-test-table/tables/00a5c8df16e2b19a6629c32ed42ef3a76d6fbac6/index.html

so I restarted the build via the travis web page like you suggested.

however I anticipate this will be an issue for almost all future builds, for example there is one missing image in your new branch's test table

http://ropensci.github.io/plotly-test-table/tables/f26e22467e4ad91b367bd5b15d9f68f304a7a4f9/density.html

From the plotly side it would be nice if @chriddyp could investigate and see why the plotly web site randomly returns an error when we ask for a png image.

From the R side I will attempt to modify the plotly-test-table code so that it checks for missing images and attempt to re-download them.

cpsievert:
httr could help on the R side. Instead of [these lines](https://github.com/ropensci/plotly-test-table/blob/gh-pages/table.R#L225-226), you could do something like

```r
library(httr)
g <- GET(plotly.png.url, write_disk(plotly.png.file))
warn_for_status(g)
# if we want to stop the build, use stop_for_status() instead
```

tdhock:
thanks for the suggestion, I will look into that.

by the way after re-building the table it seems that the missing PNGs now are true negatives (the ggplotly function stops with an error for master, but works for this branch)

chriddyp:
`warn_for_status` is great, the status code will be really helpful in debugging. my guess is that it is a timeout issue, but I'm not sure without seeing the status code. we could also put in a retry logic: if it failed, pause for 5 seconds, and then re-try.

chriddyp:
Hey @tdhock and @cpsievert - what do you guys think about merging in all the cookbook `.R` files (without additional fixes) so that we can get a broader set of tests in our test-table before we merge other PRs like https://github.com/ropensci/plotly/pull/178 and https://github.com/ropensci/plotly/pull/167 (and risk degradation!)

tdhock:
I still think it makes more sense to do one cookbook .R file at a time, to avoid having the tests on master failing.

tdhock:
by the way I think we can merge this branch right away, here is the test table

http://ropensci.github.io/plotly-test-table/tables/07af388fed4266ed8a87fb48ddb711aefcfa7655/index.html

R/ggplotly.R
```diff
@@ -271,6 +274,12 @@ gg2list <- function(p){
   trace.name.map <- c()
   for(xy in c("x","y")){
     ax.list <- list()
+    coord.lim <- p$coord$limits[[xy]]
+    if(is.numeric(coord.lim)){
+      ## TODO: maybe test for more exotic coord specification types
+      ## involving NA, Inf, etc?
```
chriddyp: could you add this as a github issue so that we can keep track of it?

R/ggplotly.R
```diff
@@ -335,8 +344,13 @@ gg2list <- function(p){
       sc <- p$scales$scales[[scale.i]]
       if(ax.list$type == "category"){
         trace.order.list[[xy]] <- sc$limits
+        if(is.character(sc$breaks)){
+          if(is.character(sc$labels)){
+            trace.name.map[sc$breaks] <- sc$labels
+          }
+          ##TODO: if(is.function(sc$labels)){
```
chriddyp: same here

chriddyp:
other than my comments, looks good to me :+1:

R/ggplotly.R
tdhock: added #184

R/ggplotly.R
tdhock: added #185

PR #174, opened 2015-03-04
Native R build on Travis

cpsievert:
This will take advantage of "officially supported" [native R builds on Travis](http://blog.travis-ci.com/2015-02-26-test-your-r-applications-on-travis-ci/)


cpsievert:
Since [`make` now requires **httr**](https://github.com/ropensci/plotly-test-table/blob/a74d065e94c92811c28c6eb8e5f91129ddfc7127/table.R#L3) we'll need to install it on Travis. For this reason, I'd suggest merging this pull request before any others (otherwise the builds will likely fail).

cpsievert:
Ah, I forgot Travis builds depend on devtools which depends on httr (as well as RCurl). I still think it's a good idea to list these in `.travis.yml` however [in case that ever changes](https://github.com/hadley/httr/issues/172).

cpsievert:
I'm ready to merge this when you are @tdhock @chriddyp

Only changes are to `.travis.yml` and `.push_test_table.sh`, so it won't be necessary to bump the package version

tdhock:
+1

tdhock:
this branch includes code for pushing log files to the plotly-test-table, which would have been helpful to see the plotly URLs in #193

PR #178, opened 2015-03-07
Toby rect

tdhock:
I implemented geom_rect conversion by treating it as a basic version of a polygon, which is transformed into a trace with mode="lines", type="scatter", fill="tozerox"


tdhock:
so with the current code this ggplot http://docs.ggplot2.org/current/geom_rect.html

becomes this plotly https://plot.ly/398/~tdhock/

so I guess fill="tozerox" is not good enough when there are multiple rects that I want to draw in 1 trace. Alex said in an email that 'we could always add a fill mode like "toself" tailored for this' @chriddyp do you think I should wait until that is implemented?

or I can implement this right now by making each rect a separate trace, as in https://plot.ly/3722/~TestBot/

jackparmer:
haha, well, we tried. 
why not do this with rectangle objects again instead of traces? the hover is also odd using traces with fill instead of an svg rectangle.

alexcjohnson:
oooookay, it's *possible* to get this effect with the existing fill modes, but it isn't fun:
https://plot.ly/~alex/1023

You need to duplicate the starting corner of each one, then leave a (non-numeric) gap between rects, then when you're done retrace your steps past all the start / end points (including more gaps so the lines don't get drawn), and for some bizarro reason (that's probably a bug but its such a hack at this point I'm not sure I care) you have to duplicate the final point. :hankey: 

If we made a fillmode "toself" tailored to making closed shapes it would:
- let you just specify each corner once, no duplication at the end
- make a new shape every time you leave a non-numeric gap, just by closing to the start of that shape, with no silly retracing at the end
- do what you expect with `line.shape='smooth'` (try that on the plot above to see what I mean - not that you would want this for rects, but I could imagine other uses where you want to smoothly outline some region, like in a contour plot)
- do something smarter with padding the axis ranges. I'm not sure whether it'd be padding on all sides or none, but it certainly wouldn't be padding on only the top and bottom, none on left and right, as you get with `fill='tozeroy'`
- maybe even do something smart with hover info and/or displayed text. Not sure about this one, then we're getting into territory that should really be its own trace type (like mosaic plots)

Anyway, this is all a ways out, it's a bigger project than I can just bang out right now. Until then,if you can stomach my hack, that's going to be the highest performance for a large number of rects, then only make a new trace when you want a new color or some other feature like new dashes as in https://plot.ly/3722/~TestBot/. Maybe also set `hoverinfo='none'` unless you *really* want people to see the corner values.

tdhock:
@alexcjohnson thanks for the example plotly. It is nice to know your hack will "be the highest performance for a large number of rects" and I think it makes sense to "only make a new trace when you want a new color" so I will go ahead and implement that.

tdhock:
after some experimentation in the plotly GUI, I found that I don't have to re-trace my steps all the way back to the first trace, but instead just back to the first point of the second trace 

https://plot.ly/~tdhock/402

tests/testthat/test-ggplot-rect.R
```diff
@@ -0,0 +1,88 @@
...
+                  0, 0)
+  expect_equal(tr$y, expected.y)
+})
+
+rect.color <- ggplot(df4, aes(xmin = x, xmax = x + 0.5, ymin = 0, ymax = 1)) +
+  geom_rect(aes(color=status))
+
+test_that('rect color', {
+  info <- expect_traces(rect.color, 2, "rect-color")
+  ## TODO: test for weird forward/backward NA pattern with 2 rects.
```
tdhock: Seems like conversion of geom_polygon and geom_rect is working fine now, but I would like to add a few more tests before merging with master.

R/trace_generation.R
```diff
@@ -322,14 +322,18 @@ toBasic <- list(
...
+    others <- g$data[!names(g$data) %in% used]
+    g$data <- with(g$data, {
+      rbind(cbind(x=xmin, y=ymin, others),
+            cbind(x=xmin, y=ymax, others),
+            cbind(x=xmax, y=ymax, others),
+            cbind(x=xmax, y=ymin, others))
+    })
+    g$geom <- "polygon"
+    g
   },
```
tdhock: geom_rect is converted to a basic geom "polygon"

R/trace_generation.R
```diff
@@ -322,14 +322,18 @@ toBasic <- list(
...
     group2NA(g, "path")
   },
-  polygon=function(g){
-    if(is.null(g$params$fill)){
-      g
-    }else if(is.na(g$params$fill)){
-      group2NA(g, "path")
-    }else{
-      g
-    }
```
tdhock: geom_polygons are now treated as a basic type so this code is no longer necessary

tests/testthat/test-ggplot-polygons.R
```diff
@@ -1,115 +1,141 @@
...
-  expect_equal(info[[1]]$name, "name1")
-  expect_equal(info[[2]]$x, c(12, 13, 13, 12, 12))
-  expect_equal(info[[2]]$y, c(10, 10, 11, 11, 10))
-  expect_equal(info[[2]]$fillcolor, toRGB("grey"))
-  expect_equal(info[[2]]$line$color, toRGB("springgreen3"))
-  expect_equal(info[[2]]$name, "name2")
-
-  expect_identical(info[[1]]$showlegend, TRUE)
-  expect_identical(info[[2]]$showlegend, TRUE)
+  info <- expect_traces(gg, 1, "black")
```
tdhock: I updated the tests to check for the NA values and retracing the first points of each group.

tdhock:
IMO This code is ready to merge with master. Please code review @mkcor @chriddyp @cpsievert

cpsievert:
@tdhock are the "ggplotly error" results for rect-* tests to be expected in the master branch?

cpsievert:
for example -- http://ropensci.github.io/plotly-test-table/tables/8ac051827b537f6bb1e51d0245f9d18f1d0bb69a/rects-black.html

tests/testthat/test-ggplot-rect.R
```diff
@@ -0,0 +1,147 @@
...
+  geom_rect()
+
+test_that('geom_rect becomes 1 trace with mode="lines" fill="tozerox"', {
+  info <- expect_traces(gg, 1, "black")
+  tr <- info$traces[[1]]
+  expect_identical(tr$fill, "tozerox")
+  expect_identical(tr$type, "scatter")
+  expect_identical(tr$mode, "lines")
+  for(xy in c("x", "y")){
+    expect_true(any(is.na(tr[[xy]])))
```
cpsievert: Just FYI, base does have `anyNA()` which is slightly faster

cpsievert:
This looks good to me :+1:

tdhock:
for reference here is the test table

http://ropensci.github.io/plotly-test-table/tables/10c8bcccd6b0f482f553dc772667339eeb060c2a/index.html

a new polygon test

http://ropensci.github.io/plotly-test-table/tables/10c8bcccd6b0f482f553dc772667339eeb060c2a/polygon-star-fill-color.html

a new rect test

http://ropensci.github.io/plotly-test-table/tables/10c8bcccd6b0f482f553dc772667339eeb060c2a/rect-fill-color.html

Issue #161, opened 2015-01-27
unsupported features from the r cookbook


chriddyp:
Here are some remaining unsupported features as taken from the great ggplot2 cookbook.

The cookbook examples are in this branch: https://github.com/ropensci/plotly/tree/add-r-cookbook-tests, in this folder: https://github.com/ropensci/plotly/tree/add-r-cookbook-tests/tests/cookbook-test-suite

- [ ] line chart with error bars throwing exception
    "Error in trace.list[[j]] : subscript out of bounds”
    
    example code: http://www.cookbook-r.com/Graphs/Plotting_means_and_error_bars_(ggplot2)/#line-graphs
    
    executable code: https://github.com/chriddyp/ggplot2-plotly-cookbook/blob/a45f2c70b7adf484e0b0eb8810a1e59e018adbb8/means_and_error_bars.R#L162-L191
    
    expected outputs: 
    http://www.cookbook-r.com/Graphs/Plotting_means_and_error_bars_(ggplot2)/line.png
    http://www.cookbook-r.com/Graphs/Plotting_means_and_error_bars_(ggplot2)/line-dodge.png
    http://www.cookbook-r.com/Graphs/Plotting_means_and_error_bars_(ggplot2)/line-dodge-ci.png
    http://www.cookbook-r.com/Graphs/Plotting_means_and_error_bars_(ggplot2)/line-dodge-black.png
    http://www.cookbook-r.com/Graphs/Plotting_means_and_error_bars_(ggplot2)/line-final.png
    
    

- [ ] custom legend labels, or trace.name
    see our cookbook examples "legends/legend title with custom labels" also "legends/line chart with custom legend"

- [ ] manual tick marks
    expected output: http://www.cookbook-r.com/Graphs/Axes_(ggplot2)/bp_y_break.25.png
    
    actual output: https://plot.ly/~ggplot2-cookbook/83/weight-vs-group/
    
    code: https://github.com/chriddyp/ggplot2-plotly-cookbook/blob/master/axes.R#L70-L71
    
    from: http://www.cookbook-r.com/Graphs/Axes_(ggplot2)/#setting-and-hiding-tick-markers
- [ ] axes tick mark labels
    see http://www.cookbook-r.com/Graphs/Axes_(ggplot2)/#setting-tick-mark-labels
    
    executable code here: https://github.com/chriddyp/ggplot2-plotly-cookbook/blob/master/axes.R
- [ ] manual range
[...]

PR #183, opened 2015-03-10
geom_smooth()

### LINKED ISSUE #179: R API: ggplot2 ggplotly():  geom_smooth() not supported
Trying to make a plot with geom_smooth() ends up without the smoothing layer.  The error message is

    In layer2traces(L, df, misc): Conversion not implemented for geom_smooth (basic geom_smooth),
    ignoring. Please open an issue with your example code at https://github.com/ropensci/plotly/issues

Code to reproduce is as follows

    raw_data = Quandl(c('WIKI/XOM', "NSE/OIL"))
    clean.quandl.name <- function(x) {
    cleaned <- x
    if (x=='Date') {} else {
        tryCatch({
                    parts <- stringr::str_split(x, " - ",n=2)
                    first.parts <- stringr::str_split(parts[[1]][[1]], "\\.",n=2)
                    cleaned <- paste(first.parts[[1]][[2]], parts[[1]][[2]], sep=".")
                    cleaned <- str_replace_all(cleaned," ","")
                    },
                 error = function(e) {cat(paste0("Err on",x,"\n"))}
                 )
    }
    stringr::str_trim(cleaned)
    }
    fixed.names <- lapply(names(raw_data),clean.quandl.name)
    renamed_data <- raw_data
    names(renamed_data) <- fixed.names
    gp <- ggplot(renamed_data, aes(x=OIL.Open, y=XOM.Open))+geom_point()+geom_smooth()
    py <- plotly()
    response <- py$ggplotly(gp)

My output looked like [this](https://plot.ly/~brianboonstra/52/xomopen-vs-oilopen/):

but should have looked like this:
unknown

cpsievert:
This will close #179.


cpsievert:
The test table looks good to me -- 

http://ropensci.github.io/plotly-test-table/tables/9b5bbf306ada7d25d2f7d778f2cb11f69fc437dd/index.html

Note the new tests -- 

http://ropensci.github.io/plotly-test-table/tables/9b5bbf306ada7d25d2f7d778f2cb11f69fc437d/smooth.html
http://ropensci.github.io/plotly-test-table/tables/9b5bbf306ada7d25d2f7d778f2cb11f69fc437dd/smooth-se-false.html

Perhaps I should add the cookbook scatterplot tests as well?

https://github.com/ropensci/plotly/blob/add-r-cookbook-tests/tests/cookbook-test-suite/scatterplots.R

On a related note, it might be worth having another checklist on #161 that tracked which of these testing files have been ported

chriddyp:
@cpsievert nice! yes, let's add the scatter plots cookbook file to this PR, as it contains a bunch of really good examples

tdhock:
great, @cpsievert can you add your name to the table in the comment I just made on #161 ?

chriddyp:
on another note, it seems like the color mapping is off: 
ggplot2: 
image
plotly:
image

cpsievert:
This appears to be mostly working now. Only caveat is the "color" legend. 

http://ropensci.github.io/plotly-test-table/tables/91e6e3d6dc473e62a6ec2b67a46d29460adc3fca/scatterplots-scale-color-hue.html

Is it possible to merge points and lines in a legend with plotly like ggplot? If so, anyone have a good example? @tdhock @chriddyp

chriddyp:
cool, looks good! we don't have the notion of merged legend items. for this case, we could "hide" the other legend items use the trace property "showlegend"

cpsievert:
Looks good to me! Ready to merge when you are @chriddyp @tdhock 

http://ropensci.github.io/plotly-test-table/tables/4ac88f0165a84d3a01957084dc8eb504b66d3d3a/index.html

chriddyp:
looks good :+1: ! @tdhock - will you do a code review too?

R/ggplotly.R
```diff
@@ -217,8 +217,24 @@ gg2list <- function(p){
...
+      trace1 <- if (isTRUE(L$stat_params$se == FALSE)) {
+        NULL
+      } else {
+        layer2traces(L, df, misc)
+      }
+      misc$smoothRibbon <- FALSE
+      # always draw the line
+      misc$smoothLine <- TRUE
+      trace2 <- layer2traces(L, df, misc)
+      trace2 <- lapply(trace2, function(x) { x$showlegend <- FALSE; x })
```
tdhock: I think it is confusing to use lapply with an anonymous function here where you could have used a for loop.

R/ggplotly.R
```diff
@@ -217,8 +217,24 @@ gg2list <- function(p){
     }
     
     # This extracts essential info for this geom/layer.
-    traces <- layer2traces(L, df, misc)
-
+    if (L$geom$objname == "smooth") {
```
tdhock: This geom_smooth-dependent code should be moved inside of layer2traces.

cpsievert: The code in trace_generation.R seems to assume a layer consists of one geom (is that right?), but a `geom_smooth()` is really `geom_path() + geom_ribbon()`. I suppose I could add `smooth` (instead of `smoothRibbon` & `smoothLine`) to these lists, but the downside is that we essentially have to copy/paste over `ribbon` and `line` into this spot (so, if we happen to fix/change `ribbon` or `line`, those changes won't propagate to `smooth`). Would you prefer to take this approach instead? Or maybe there is a better way?

tdhock: Well actually the approach with smoothRibbon and smoothLine is fine with me, but I just think it will be easier to maintain in the future if we put all the geom-specific code inside of layer2traces (rather than being inside of gg2list as it is now).

cpsievert:
I managed to avoid geom specific code in ggplotly.R via a recursive call to `layer2traces()`. The test table looks good. Let me know what you think @tdhock 

http://ropensci.github.io/plotly-test-table/tables/a97ba2b6b71e3f335099d9e73cc13ea18a13e14e/index.html

tdhock:
+1 thanks for making that change!

Everything looks good to me, in the code and on the test table.

R/ggplotly.R
mkcor: @tdhock What?! I would say that, in R, you want to use *ply functions (in order to avoid for loops) as much as possible!!

Issue #190, opened 2015-03-18
Ribbons in geom_smooth 


cpsievert:
A user tried to recreate the following example in plotly. 

```r
p <- (qplot(Date, Score, colour=Program, data=x)
      +stat_smooth(method=lm)+facet_wrap(~Name)
      +scale_x_date(labels=date_format("%b/%y"))
      +scale_color_fivethirtyeight()+theme_fivethirtyeight())
```

![asada vs yuna](https://cloud.githubusercontent.com/assets/1365941/6712007/61cbd258-cd58-11e4-8bad-a1504b0ef455.png)

However, [the plotly result](https://plot.ly/~sungho.hong/45) is:

![](https://plot.ly/~sungho.hong/45.png)

This is clearly a bug, but before I fix it, it'd be nice to have a discussion on how best to implement the ribbon in `geom_smooth()`.

Alex mentioned that each ribbon should should be 2 traces, so that upon hover in compare mode, we can compare the top and bottom. I definitely agree that would be useful. My question however is: [why isn't `geom_ribbon()` implemented this way](https://plot.ly/~cpsievert/245)? 

The current approach is basically to treat `geom_smooth()` as `geom_line() + geom_ribbon()`. So if we re-implement `geom_ribbon()` as 2 traces, that change should propagate to `geom_smooth()`.

Thoughts @tdhock @chriddyp @mkcor?

tdhock:
geom_ribbon has both aes(fill) and aes(color), so if we want to be able to translate both of those to plotly I think we should treat it as a basic polygon (like geom_rect, see #178). To support comparing top and bottom values, I propose that we add an invisible error bar for each ribbon.

PR #193, opened 2015-03-18
Carson ribbon

cpsievert:
This pull request re-implements ribbon as a basic polygon (in a similar fashion to [rect](https://github.com/ropensci/plotly/pull/178)).

One major benefit of doing so is that `geom_ribbon()` (and consequently `geom_smooth()`)  will respect fill/colour/group aesthetics.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: plotly
 Type: Package
 Title: Interactive, publication-quality graphs online.
-Version: 0.5.26
+Version: 0.5.27
 Authors@R: c(person("Chris", "Parmer", role = c("aut", "cre"),
     email = "chris@plot.ly"),
     person("Scott", "Chamberlain", role = "aut",
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,7 @@
+0.5.27 -- 19 Mar 2015
+
+Reimplement geom_ribbon as a basic polygon. Fix #191. Fix #192.
+
 0.5.26 -- 18 Mar 2015
 
 Implemented geom_rect #178
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -598,13 +598,36 @@ gg2list <- function(p){
       layout$showlegend <- FALSE
     }
   }
-
+  
+  # avoid redundant legends
+  fills <- lapply(trace.list, function(x) paste0(x$name, "-", x$fillcolor))
+  linez <- lapply(trace.list, function(x) paste0(x$name, "-", x$line$color))
+  marks <- lapply(trace.list, function(x) paste0(x$name, "-", x$marker$color))
+  fill_set <- unlist(fills)
+  line_set <- unlist(linez)
+  mark_set <- unlist(marks)
+  legend_intersect <- function(x, y) {
+    i <- intersect(x, y)
+    # restrict intersection to valid legend entries
+    i[grepl("-rgb[a]?\\(", i)]
+  }
+  # if there is a mark & line legend, get rid of line
+  t1 <- line_set %in% legend_intersect(mark_set, line_set)
+  # if there is a mark & fill legend, get rid of fill
+  t2 <- fill_set %in% legend_intersect(mark_set, fill_set)
+  # if there is a line & fill legend, get rid of fill
+  t3 <- fill_set %in% legend_intersect(line_set, fill_set)
+  t <- t1 | t2 | t3
+  for (m in seq_along(trace.list)) 
+    if (trace.list[[m]]$showlegend && t[m]) 
+      trace.list[[m]]$showlegend <- FALSE
+  
   # Only show a legend title if there is at least 1 trace with
   # showlegend=TRUE.
   trace.showlegend <- sapply(trace.list, "[[", "showlegend")
   if (any(trace.showlegend) && layout$showlegend && length(p$data)) {
     # Retrieve legend title
-    legend.elements <- sapply(traces, "[[", "name")
+    legend.elements <- unlist(sapply(traces, "[[", "name"))
     legend.title <- ""
     for (i in 1:ncol(p$data)) {
       if (all(legend.elements %in% unique(p$data[, i])))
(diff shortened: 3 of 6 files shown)


cpsievert:
It's worth noting that I haven't (yet) added the invisible errorbars ([as suggested by Toby](https://github.com/ropensci/plotly/issues/190#issuecomment-83040426)) to aid comparison of upper and lower bounds of ribbons upon hover. If there are multiple groups, that could get quite busy, so my suggestion would be to only draw the invisible errorbars when there is 1 group. Does anyone object to that idea? If not, I'll likely implement that in a separate pull request.

cpsievert:
Note the new tests --

http://ropensci.github.io/plotly-test-table/tables/39b254657b53c4d5fba34d07c6dde38db64c3ded/ribbon-group.html

http://ropensci.github.io/plotly-test-table/tables/39b254657b53c4d5fba34d07c6dde38db64c3ded/ribbon-colour.html

http://ropensci.github.io/plotly-test-table/tables/39b254657b53c4d5fba34d07c6dde38db64c3ded/ribbon-fill.html

http://ropensci.github.io/plotly-test-table/tables/39b254657b53c4d5fba34d07c6dde38db64c3ded/smooth-group.html

http://ropensci.github.io/plotly-test-table/tables/39b254657b53c4d5fba34d07c6dde38db64c3ded/smooth-colour.html

http://ropensci.github.io/plotly-test-table/tables/39b254657b53c4d5fba34d07c6dde38db64c3ded/smooth-fill.html

The smoother legends still need a bit of work...hopefully I can fix that tomorrow.

tdhock:
looks good to me. did you add any tests for ribbons and facets?

cpsievert:
No, but good idea. Do you think I should put them in a separate test file?

tdhock:
same file is fine with me.

cpsievert:
Here is a test for facetted smooths -- 

http://ropensci.github.io/plotly-test-table/tables/aaf302ed8f9fb55b52f90b156425cb9f1124f52e/smooth-facet.html

Unfortunately [my attempt](https://github.com/ropensci/plotly/commit/b50a96952d482c3322a235928970cb424c986ed1) to avoid redundant legend entries breaks this plot in a weird way (I'm not sure why) -- 

http://ropensci.github.io/plotly-test-table/tables/154b7790508a5da192297f191a10e2c94fb337f7/path-line-symbols.html

Does anyone have any other thoughts/opinions on how to best avoid redundant legend entries?

cpsievert:
The test table looks good to me. Ready to merge when you are @tdhock @chriddyp @mkcor 

http://ropensci.github.io/plotly-test-table/tables/2e26cc0e393426fd66c5ae47440c4e1657c343ec/index.html

tdhock:
this plot still has multiple legend entries ... can you add a test for that? and then fix it?

http://ropensci.github.io/plotly-test-table/tables/2e26cc0e393426fd66c5ae47440c4e1657c343ec/smooth-facet.html

cpsievert:
At first I was thinking that was the correct behavior since the legend entries have different alpha levels; however, the more I think about it, in most cases, the differences in alpha should be ignored (especially if the entry names are the same). Anyway, I go ahead and make sure this becomes one legend.

cpsievert:
OK, legend entry comparison will now ignore the alpha level, so we get the desired result here -- 

http://ropensci.github.io/plotly-test-table/tables/a13e92794153452ae16252c6799dba7983462aae/smooth-facet.html

The rest of the table looks good to me. Go ahead and merge when ready @tdhock @chriddyp @mkcor 

http://ropensci.github.io/plotly-test-table/tables/a13e92794153452ae16252c6799dba7983462aae/

chriddyp:
hey sorry just checking in on this. I don't understand why fill/colour/group can't be done with multiple traces and fill?

chriddyp:
is it possible to see the json or url for this test? http://ropensci.github.io/plotly-test-table/tables/a13e92794153452ae16252c6799dba7983462aae/smooth-fill2.html
image

is the fill a polygon or a trace?

tdhock:
when we make the test table we write a log file for every download attempt https://github.com/ropensci/plotly-test-table/blob/gh-pages/table.R#L250 

for your example there is supposed to be this log file http://ropensci.github.io/plotly-test-table/data/a13e92794153452ae16252c6799dba7983462aae/smooth-fill2.log which should contain the plotly URL

but it is not found. why are logs not getting pushed to the plotly-test-table repos?

cpsievert:
The test table is slightly broken right now, so let's postpone this until that's fixed. Follow the discussion [here](https://github.com/ropensci/plotly-test-table/issues/7)

tdhock:
i fixed the bugs and re-built the test table, voilà

http://ropensci.github.io/plotly-test-table/tables/ef301b0e7c602b1cb1484f73426ac30dcbf24983/big.html

cpsievert:
Nice! Thanks @tdhock! 

I merged that pull request which makes logs available, so @chriddyp, you can get [the url](https://plot.ly/~TestBot/7776/price-vs-carat/) for that example via 

http://ropensci.github.io/plotly-test-table/data/ef301b0e7c602b1cb1484f73426ac30dcbf24983/smooth-fill2.log

cpsievert:
@chriddyp multiple traces will be generated if the colour/fill aesthetic is specified, but not for group. Do you want multiple traces so that we have more informative tooltips? I think that should be possible...I'm just using the default functionality for polygons (these are actually ribbons -- a special case of a polygon)

tdhock:
for the interactivity it would be nice if the popup was only displayed for x-values with black dots, so then users could compare the observed data value with the mean data value.

that could be achieved if we evaluate the smooth line only on the grid of x-values that are present in the data set (rather than a uniformly spaced grid which I guess is what the code currently does).

cpsievert:
Good idea @tdhock. Unfortunately I'm swamped with other stuff this week. This might have to wait until next week.

chriddyp:
oh, nvm, when you guys were saying "basic polygon" I thought you meant Plotly's new basic polygon shapes (https://plot.ly/~etpinard/1647/plotly-shapes/).

This looks great to me! :dancer:

cpsievert:
Great, I'm keen on merging this since it's such a big bug fix. We can work out the details of comparison mode in another pull request. Are you OK with that @tdhock?

tdhock:
fine with me +1

PR #196, opened 2015-03-31
Toby cookbook lines

tdhock:
I added some tests based on https://github.com/ropensci/plotly/blob/add-r-cookbook-tests/tests/cookbook-test-suite/lines.R

one test doesn't pass, and I'm working on fixing it.


tdhock:
fixing the one test that failed turned out to more complicated than I imagined.

I made some non-trivial changes which I will try to explain inline.

please code review @mkcor @chriddyp @cpsievert

R/ggplotly.R
```diff
@@ -97,8 +97,10 @@ gg2list <- function(p){
   # worry about combining global and layer-specific aes/data later.
   for(layer.i in seq_along(p$layers)) {
     layer.aes <- p$layers[[layer.i]]$mapping
-    to.copy <- names(p$mapping)[!names(p$mapping) %in% names(layer.aes)]
-    layer.aes[to.copy] <- p$mapping[to.copy]
+    if(p$layers[[layer.i]]$inherit.aes){
+      to.copy <- names(p$mapping)[!names(p$mapping) %in% names(layer.aes)]
+      layer.aes[to.copy] <- p$mapping[to.copy]
+    }
     mark.names <- markUnique[markUnique %in% names(layer.aes)]
```
tdhock: some geoms (like hline and vline) need to ignore the global aes mapping.

R/ggplotly.R
```diff
@@ -107,19 +109,73 @@ gg2list <- function(p){
...
+      ranks <- seq_along(br)
+      names(ranks) <- br
+      misc$breaks[[a.vec]] <- ranks
+    }
+    ## store if this is a reverse scale so we can undo that later.
+    if(is.character(sc$trans$name)){
+      misc$trans[sc$aesthetics] <- sc$trans$name
+    }
+  }
+  reverse.aes <- names(misc$trans)[misc$trans=="reverse"]
```
tdhock: this whole block which creates misc was inside the for(layer.i) block below, which is potentially confusing, since misc contains plot-specific, not layer-specific, information.

R/ggplotly.R
```diff
@@ -107,19 +109,73 @@ gg2list <- function(p){
...
+    }else{
+      ## for categorical variables on the axes, panel$ranges info is
+      ## meaningless.
+      name.name <- paste0(xy, ".name")
+      sapply(built$data, function(df){
+        paste(df[[name.name]])
+      })
+    }
+    ranges.list[[xy]] <- range(range.values)
+  }
```
tdhock: getting global ranges is more complicated now that we respect inherit.aes

R/ggplotly.R
```diff
@@ -203,21 +215,24 @@ gg2list <- function(p){
     for (a in replace.aes) {
       prestats[[a]] <- -1 * prestats[[a]]
     }
-    misc$prestats.data <-
+    L$prestats.data <-
```
tdhock: prestats.data is layer-specific (misc is plot-specific)

tdhock:
http://ropensci.github.io/plotly-test-table/tables/8e68d8abfe1137ac265e8c80f92502b3fa3c7892/big.html shows some untested regressions. I will add tests and make the fixes.

tdhock:
still need to fix these test failures

```
1. Failure (@test-cookbook-lines.R#215): Add colored lines for the mean xval of each group 
tr$showlegend isn't false

2. Failure (@test-cookbook-lines.R#215): Add colored lines for the mean xval of each group 
tr$showlegend isn't false

3. Failure (@test-cookbook-lines.R#270): geom_line -> 2 more traces ------------
tr$showlegend is not identical to expected. Differences: 
1 element mismatch

4. Failure (@test-cookbook-lines.R#270): geom_line -> 2 more traces ------------
tr$showlegend is not identical to expected. Differences: 
1 element mismatch

```

tdhock:
Also I'm wondering if anybody has ideas about how to best translate this to plotly?

http://ropensci.github.io/plotly-test-table/tables/8e68d8abfe1137ac265e8c80f92502b3fa3c7892/cookbook-lines-bar-dodge-color-error.html

tdhock:
also this one? 

http://ropensci.github.io/plotly-test-table/tables/8e68d8abfe1137ac265e8c80f92502b3fa3c7892/cookbook-lines-bar-dodge-color-err4.html

tdhock:
all the tests I wrote pass now, but there are still the visual problems I mentioned above which I am not sure how to convert to plotlys. Any ideas?

chriddyp:
It seems like those bar lines are a ggplot2 error bars hack.

```
# Draw separate hlines for each bar. First add another column to dat
dat$hline <- c(9,12)
dat
#>        cond result hline
#> 1   control   10.0     9
#> 2 treatment   11.5    12

# Need to re-specify bp, because the data has changed
bp <- ggplot(dat, aes(x=cond, y=result)) +
    geom_bar(position=position_dodge(), stat="identity")

# Draw with separate lines for each bar
bp + geom_errorbar(aes(y=hline, ymax=hline, ymin=hline), colour="#AA0000")

# Make the lines narrower
bp + geom_errorbar(width=0.5, aes(y=hline, ymax=hline, ymin=hline), colour="#AA0000")
```
image
image

At some point, `ggplotly` was merging `geom_errors` with `geom_bar` and `geom_line` traces if they shared the same `x` and `y` data. In this case, these `geom`s don't share `x` and `y` data, so either we keep them attached to invisible bar traces or we convert them to plotly's new `layout.shape.line` objects.

**Option 1 - Attaching them to invisible bar traces**
*Pros*
- position is really straightforward
- ggplot2 has width relative to bar width (whereas we just have the pixel width). If we had relative widths, specifying `width` would be really straightforward too (but we realistically won't do this for months)

*Cons*
- Semantically it's weird to have floating error bars
- Since we do width in pixels right now, it'll be a lot of work to get this width exactly right

**Option 2 - Making them separate line objects**
*Pros*
- Semantically it seems more like what the user wanted 
- Computing width is possible with data coordinates and `bargap` and `bargroupgap` [computations](https://github.com/plotly/streambed/blob/master/shelly/plotlyjs/static/plotlyjs/src/bars.js#L220-L284)

*Cons*
- Computing position will be a little annoying with `bargap` and `bargroupgap`
- There won't be a legend entry for these lines

Here's an example with lines, and `bargap` and `bargroupgap` set to `0`
Bar chart with shape lines
```python
{
    'data': [
        {
            'x': ['Cats', 'Dogs'],
            'y': [4, 6],
            'type': 'bar',
            'name': 'Series 1'
        },
        {
            'x': ['Cats', 'Dogs'],
            'y': [5, 7],
            'type': 'bar',
            'name': 'Series 2'
        }
    ],
    'layout': {
        'bargap': 0,
        'shapes': [
            {
                'type': 'line',
                'x0': -0.5,
                'x1': 0.5,
                'y0': 4.5,
                'y1': 4.5,
                'line': {
                    'color': 'blue',
                    'width': 4,
                    'dash': 'dash'
                }
            },                
            {
                'type': 'line',
                'x0': 0.5,
                'x1': 1.5,
                'y0': 5.5,
                'y1': 5.5,
                'line': {
                    'color': 'red',
                    'width': 4,
                    'dash': 'dot'
                }
            },
            {
                'type': 'line',
                'x0': 0.5,
                'x1': 1,
                'y0': 6.5,
                'y1': 6.5,
                'line': {
                    'color': 'green',
                    'width': 4,
                    'dash': 'dashdot'
                }
            },
            {
                'type': 'line',
                'x0': 0,
                'x1': 0.5,
                'y0': 7.0,
                'y1': 7.0,
                'line': {
                    'color': 'orange',
                    'width': 4,
                    'dash': 'solid'
                }                    
            }                
        ]
    }
}
```

chriddyp:
Also looks like there is a regression with legend hide/show
image

R/trace_generation.R
```diff
@@ -358,8 +363,7 @@ toBasic <- list(
     g
   },
   bar=function(g) {
-    if (any(is.na(g$prestats.data$x)))
-      g$prestats.data$x <- g$prestats.data$x.name
+    ## TODO: why is this here? bar is a basic geom.
```
mkcor: I guess a 'git blame' would help tracing this back.
I agree it looks non-robust.  I remember doing this so that popular use cases convert well...

mkcor:
Hi @tdhock sorry for not being on R lately... Thanks for this work. I believe all the `TODO` comments should belong to a separate PR: I mean, it's a discussing worth having, but it shouldn't pollute something else you're fixing now.
Also, can you please follow the style guide? Again, comments start with one `#`.

tdhock:
@chriddyp thanks for your analysis about how to convert this very strange geom_errorbar to plotly.

Rather than spending more time debating which one of those two options we should implement, I suggest we wait until users encounter problems. And then we can choose the implementation as a function of what users expect. Is that OK?

@mkcor I moved the TODOs from comments to an issue #203 and I used single # for comments.

If this is fine with you (and @cpsievert ) then I will proceed to merge with master.

tdhock:
by the way here is the complete visual test table. I did not notice any regressions.

http://ropensci.github.io/plotly-test-table/tables/96f0d4073aabf455731b0db3775c41d9830a24e6/index.html

chriddyp:
@tdhock - sounds good and looks good!

R/ggplotly.R
```diff
@@ -107,19 +110,79 @@ gg2list <- function(p){
...
-  ggxmax <- max(sapply(xrange, max))
-  # Extract y.range
-  yrange <- sapply(ggranges, `[[`, "y.range", simplify=FALSE, USE.NAMES=FALSE)
-  ggymin <- min(sapply(yrange, min))
-  ggymax <- max(sapply(yrange, max))
+  # Get global ranges now because we need some of its info in layer2traces
+  ranges.list <- list()
+  for(xy in c("x", "y")){
+    use.ranges <-
+      misc$is.continuous[[xy]] |
```
cpsievert: Should probably use `||` here , not `|`

cpsievert:
This isn't super important, but it'd be nice to see the color of the lines match the respective group here

http://ropensci.github.io/plotly-test-table/tables/96f0d4073aabf455731b0db3775c41d9830a24e6/cookbook-lines-bar-dodge-color-err4.html

Also, maybe the default color should be black?

http://ropensci.github.io/plotly-test-table/tables/96f0d4073aabf455731b0db3775c41d9830a24e6/cookbook-lines-bar-dodge-color-error.html

mkcor:
@cpsievert I think the errorbars are green in the `plotly` plot because they correspond to the third trace, and green is the third default colour in plotly... So maybe we want to be more manual about it and convert the colour (black being the fallback).

Funnily enough, we haven't bothered (yet) with the first default colour, which is a black in `ggplot` and blue in `plotly` (e.g., http://ropensci.github.io/plotly-test-table/tables/96f0d4073aabf455731b0db3775c41d9830a24e6/cookbook-lines-basic-bar.html).

cpsievert:
@mkcor good to know, I've noticed the default color/fill are different on several tests. If you them to match, just say the word and I'll take care of it :)

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/60974677 

 On TravisCI, commit 9ab52e198a663653cf039add62563df6a8becdc6 was successfully merged with d9ecaf43fa75eeef0142cef93361fdd1e936d227 (master) to create 9abb9dea695bd9d8320629392c2b2f5be87e1b49. A visual testing table comparing d9ecaf43fa75eeef0142cef93361fdd1e936d227 with 9abb9dea695bd9d8320629392c2b2f5be87e1b49 can be found here:
 http://ropensci.github.io/plotly-test-table/tables/9abb9dea695bd9d8320629392c2b2f5be87e1b49/index.html

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/60974677 

 On TravisCI, commit 9ab52e198a663653cf039add62563df6a8becdc6 was successfully merged with d9ecaf43fa75eeef0142cef93361fdd1e936d227 (master) to create 9abb9dea695bd9d8320629392c2b2f5be87e1b49. A visual testing table comparing d9ecaf43fa75eeef0142cef93361fdd1e936d227 with 9abb9dea695bd9d8320629392c2b2f5be87e1b49 can be found here:
 http://ropensci.github.io/plotly-test-table/tables/9abb9dea695bd9d8320629392c2b2f5be87e1b49/index.html

cpsievert:
@chriddyp @mkcor this looks ready to merge now. I can work on a more general fix on default colors/sizing in another pull request.

mkcor:
@cpsievert Okay, looking now. I'm confused though, because files `DESCRIPTION` and `NEWS` shouldn't be edited before getting the +1 (since you can't tell which day you'll be getting that +1).

cpsievert:
Yea, they were edited before I merged with master, so I had to resolve those conflicts...

R/ggplotly.R
```diff
@@ -628,12 +643,17 @@ gg2list <- function(p){
...
   }
+  layout$showlegend <- if(length(legends.present) == 0) FALSE else TRUE
   for(a in legends.present){
     if(is.hidden(p$guides[[a]])){
       layout$showlegend <- FALSE
     }
   }
-  
+  # Legend hiding from theme.
+  if(theme.pars$legend.position=="none"){
```
mkcor: Spaces *at least* around `==` and between `) {`.

R/trace_generation.R
```diff
@@ -46,7 +46,7 @@ layer2traces <- function(l, d, misc) {
     g$geom <- "bar"
     bargap <- 0
   }
-  
+
```
mkcor: No, I'm pretty sure this indentation is from RStudio.

R/trace_generation.R
```diff
@@ -131,7 +136,7 @@ layer2traces <- function(l, d, misc) {
   
   # symbol=circle,square,diamond,cross,x,
   # triangle-up,triangle-down,triangle-left,triangle-right
-  
+
```
mkcor: Same here.

mkcor:
@cpsievert [default color/fill] No, thanks. Blue looks better, plus we've never heard anybody complain about it. ;)

mkcor:
@cpsievert Okay... But "5 May 2015" lies in the future so... How can it be something we would want or something we would get from somewhere?

mkcor:
I'll pull this branch and do the stylistic fixes.

mkcor:
@cpsievert Let either of us +1 and the other will merge...

cpsievert:
FYI, I think most of those style changes are due to the "Strip trailing horizontal space" option (which I'll turn on for this project)

screen shot 2015-05-05 at 4 50 46 pm

mkcor:
You mean, turn off (unchecking the box)? You're right, I always left this box unchecked so I assumed it was RStudio's default. We've had this conversation in the past: https://github.com/ropensci/plotly/pull/151#discussion_r22366138 and subsequent comments.

cpsievert:
Oh, yes, sorry, I must've not saved those files, so they were still left over from Toby's commits.

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/61377823 

 On TravisCI, commit 4cfedf95b5a16632128660af1868f2b700779ed6 was successfully merged with ae947fcf0b0fb5669e6d3b251f38e7b49798813f (master) to create 0832b38fdce185d6fdfb28750cfcc2de2ceab571. A visual testing table comparing ae947fcf0b0fb5669e6d3b251f38e7b49798813f with 0832b38fdce185d6fdfb28750cfcc2de2ceab571 can be found here:
 http://ropensci.github.io/plotly-test-table/tables/0832b38fdce185d6fdfb28750cfcc2de2ceab571/index.html

cpsievert:
Looks good! +1

cpsievert:
I had to restart one of the builds due to a silly Travis error, but this is good to go, please merge @mkcor :+1:

Issue #199, opened 2015-04-09
geom_density() as filled area chart


cpsievert:
To me, `geom_density()` should translate to a [filled area chart](https://plot.ly/javascript-graphing-library/filled-area-plots/). Right now it's translated to a [histogram](https://plot.ly/javascript-graphing-library/histograms/) (with autobinning). As a result, most [arguments to `stat_density()`](http://docs.ggplot2.org/current/stat_density.html) are ignored and `geom_density() + geom_histogram()` doesn't seem to make sense. Consider the following:

```r
p <- ggplot(mtcars, aes(wt)) + geom_density() + geom_histogram(aes(y = ..density..))
```

![rplot](https://cloud.githubusercontent.com/assets/1365941/7061027/480160f4-de51-11e4-8630-777f03402d3f.png)

This actually gives me an error:

```r
ggplotly(p)
Error in if (is.character(new.mode) && !is.na(new.mode)) { : 
  missing value where TRUE/FALSE needed
```

Any thoughts/objections @tdhock @chriddyp @mkcor?

tdhock:
seems fine with me .. are there any tests that would need to be changed?

cpsievert:
Yea, test-ggplot-density.R would have to change.

PR #200, opened 2015-04-10
Let ggplot handle histogran binning. Fix #198

cpsievert:
This pull request started out as a fix for #198, but I quickly realized the implementation of `geom_bar()`/`geom_histogram()` was suboptimal for a couple reasons 

1. The binning for histograms was done by plotly. Not only is this default binning different, but that also means a number of [arguments in stat_bin](http://docs.ggplot2.org/current/stat_bin.html) are ignored. If the goal is to have a correct mapping from ggplot2 to plotly, we should let ggplot2 handle the binning, then use a type="bar" and bargap=0.

2. [This thinking](https://github.com/ropensci/plotly/blob/a9daa28a2069cf437059e642c227b082fa882f66/R/trace_generation.R#L47-L54) lead to false warnings. For example, `ggplot(mtcars, aes(factor(vs)))+geom_bar()` would throw a warning, but it shouldn’t.

PR #202, opened 2015-04-13
geom_density() as filled area chart

cpsievert:
See #199.

### DIFF
--- a/tests/testthat/test-ggplot-density.R
+++ b/tests/testthat/test-ggplot-density.R
@@ -1,6 +1,6 @@
 context("Probability density")
 
-expect_traces <- function(gg, n.traces, name){
+expect_traces <- function(gg, n.traces, name) {
   stopifnot(is.ggplot(gg))
   stopifnot(is.numeric(n.traces))
   save_outputs(gg, paste0("density-", name))
--- a/tests/testthat/test-ggplot-density.R
+++ b/tests/testthat/test-ggplot-density.R
@@ -21,16 +21,16 @@ base <- ggplot(mtcars, aes(wt))
 test_that("geom_density() is translated to area chart", {
   info <- expect_traces(base + geom_density(), 1, "simple")
   tr <- info$traces[[1]]
-  expect_identical(tr[[1]]$type, "scatter")
-  expect_identical(tr[[1]]$fill, "tozeroy")
-  expect_identical(tr[[1]]$fillcolor, "rgba(51,51,51,0)")
+  expect_identical(tr$type, "scatter")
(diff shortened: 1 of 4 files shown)


cpsievert:
Note that 261244ada631b2dcf9d603989572621c65445160 is really the only commit relevant to this PR (the rest of the commits are carried over from #200)

cpsievert:
New tests --

http://ropensci.github.io/plotly-test-table/tables/4659669a2ea26bb8537cfc70b6e714afd6f62094/density-simple.html

http://ropensci.github.io/plotly-test-table/tables/4659669a2ea26bb8537cfc70b6e714afd6f62094/density-fill.html

http://ropensci.github.io/plotly-test-table/tables/4659669a2ea26bb8537cfc70b6e714afd6f62094/density-color.html

Something strange is happening with the density-fill example. If I click on "fork and edit", it takes me to another page that renders the plot correctly 

screen shot 2015-04-13 at 4 50 33 pm

Could this be some sort of bug with plotly itself @chriddyp @mkcor ?

tdhock:
Nice tests, and yeah that it weird about the difference between the rendered plotlys.

@alexcjohnson do you have any idea why this same plotly is rendered differently in these two cases?

cpsievert:
I don't think I can do anything to fix the problem above, so I'd prefer to merge, then post that rendering issue somewhere more appropriate. Any ideas on where that might be @chriddyp @mkcor @tdhock?

Please review and +1 when you get a chance.

alexcjohnson:
Ah, apologies for not looking at this before. I just did tried "fork and edit" and didn't see a change... but I can imagine where it could come from. The issue is the `null` at the end of the x array in the first trace. You're right that this *shouldn't* break it like this, so I'll open an issue for it on the plotly side - but I also imagine it's not quite what you had in mind here...

cpsievert:
Nice, thanks for catching this @alexcjohnson!

I'll push a fix shortly.

cpsievert:
Everything looks good now!

http://ropensci.github.io/plotly-test-table/tables/4a336c65d8cdc0a4f974912a4e93d45e1852e003/index.html

tests/testthat/test-ggplot-density.R
```diff
@@ -1,15 +1,45 @@
 context("Probability density")
 
+expect_traces <- function(gg, n.traces, name){
```
mkcor: Space after closing `)`.

mkcor:
Nice work, @cpsievert !
I only spotted a missing whitespace. Up to that, that's a +1 from me.

PR #205, opened 2015-04-16
Include license text in LICENSE file

mkcor:
Reference: http://opensource.org/licenses/MIT

/cc @tdhock @chriddyp

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,7 +18,7 @@ Authors@R: c(person("Chris", "Parmer", role = c("aut", "cre"),
     email="cpsievert1@gmail.com"))
 Author: Chris Parmer
 Maintainer: Marianne Corvellec <marianne@plot.ly>
-License: MIT + file LICENSE
+License: MIT
 Description: An interface to plotly's online graphing tools with desktop R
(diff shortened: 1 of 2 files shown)


cpsievert:
@mkcor `MIT + file LICENSE` is the format that CRAN prefers. See http://r-pkgs.had.co.nz/description.html#license

mkcor:
@cpsievert Thanks! Oh, I didn't look into R culture, I have to admit. I wanted to make things more standard/clear following comments by @tdhock but it looks like our current license info already complies with the CRAN standard... So maybe I should just close this PR. Sorry for the noise.
Toby, what do you say?

cpsievert:
No problem @mkcor, it's a silly convention. BTW, is the goal to submit to CRAN? The NOTEs/WARNINGs I see from R CMD check look fixable. The one potential problem is the use of `ggplot2:::plot_theme`, but we could just copy/paste avoid CRAN yelling at us...

mkcor:
@cpsievert Short term, not really.  But you are right, it shouldn't be too much work.  There's also the `IRdisplay` dependency.

Issue #206, opened 2015-04-16
Define parameters for `plotly()`


sckott:
Maybe i'm missing something, but could parameters defs be added for the parameters in a call to `x$ggplotly()` ?  eg.. when I look at the function definition, 

```r
function (gg = last_plot(), kwargs = list(filename = NULL, fileopt = NULL, 
    width = NULL, height = NULL), session = "interactive") 
...
```

e.g., `gg`, `kwargs`, and `session` don't seem to be documented

a user commented about this in the guest blog post on our site https://ropensci.org/blog/2014/04/17/plotly/#comment-1971263190

mkcor:
@sckott They are not documented in a docstring indeed. There are several examples around the web... But I agree this should be part of the package. How would you go about writing a docstring when the function is a method living under `plotly <- function(username=NULL, key=NULL, base_url=NULL) { ... }`?

sckott:
hmm, not sure. Though the way you set up the function, [R6](https://github.com/wch/R6) seems like a perfect fit . It sounds like there isn't yet, but will be soon support for documenting R6 in roxygen2, eg., see  https://github.com/klutometis/roxygen/issues/306

There are some workaround discussed in the `R6` issues here https://github.com/wch/R6/issues/3

cpsievert:
Fixed as of version 1.0

sckott:
:+1:

Issue #40, opened 2014-05-23
Proposed re-design


chriddyp:
With the configuration files we can re-architect the module to be a little more "R"-y. I propose the following interface:

### Before

```
py = plotly(username, key)
trace0 = list(x=...y=...)
py$plotly(trace0)
```

### After

```
plotly:::set_credentials(username, key)
plotly:::plot(trace0)
```
Or, if you prefer to send your credentials on a per-call basis (e.g. you're using the R demo account):

```
plotly:::plot(trace0, username=..., key=....)
```

And similarly for the other functions:

```
plotly:::ggplotly(..., plotly_options)
plotly:::plot(..., plotly_options)
plotly:::export_image(..., plotly_options)
plotly:::get_figure(..., plotly_options)
```

### With additional functionality ...

```
figure = plotly:::get_figure(file_owner, file_id, plotly_options) # returns a list of lists of lists of ... (plotly's JSON graph JSON)

data = plotly:::get_data(figure) # doesn't make a request to plotly, it just removes strips keys out of the named list
```

### A few questions I have (because I'm not an R expert)

1 - Is it the R convention to have nested modules/functions?

e.g. in Python we do: 
```
import plotly.plotly as py

# all the functions in py, call the plotly servers
py.plot(...)
py.iplot(...)
py.get_figure(...)
py.image.save_as(...)

# other functions, like tools, are in a separate namespace:

```python
import plotly.tools as tls
tls.embed(figure)
```

Do we do a similar thing in R? Or do we keep all the functions at the same "level", e.g.

This:

```
plotly:::embed(figure)
```
or this:

```
plotly:::tools:::embed(figure) #??
```

cpsievert:
I agree with @chriddyp that the current interface is confusing. As I understand it, the main reason for having a function return a list of functions is to enable a mutable state. As far I can tell, that isn't necessary here.

In addition to avoiding [permission issues](https://github.com/ropensci/plotly/issues/204) when reading/writing credentials to disk, if we kept track of them via environment variables, you could still specify non-default credential on a per-call basis:

```r
plotly <- function(p, username, key, ...) {
  if (missing(username)) username <- Sys.getenv("PLOTLY_USER")
  if (missing(username)) username <- Sys.getenv("PLOTLY_KEY")

  ....
}
```

Then you could just do

```r
# use default credentials
plotly::plotly(qplot(1:10))
```

or 

```r
plotly::plotly(qplot(1:10), "my_username", "my_key")
```

[...]

PR #210, opened 2015-04-28
Let gg2list return a figure object

mkcor:
`gg2list()` returns a list of lists which is converted into the final JSON file by `ggplotly()`.

It makes a lot more sense in the plotly framework to have `gg2list()` return a figure object, i.e., with dimensions `data` and `layout`. This way, users can apply it to a ggplot2 plot, say, `gg`:
```r
fig <- gg2list(gg)
# Get/set layout attributes
fig$layout
# Get/set data attributes
fig$data
# And only then make the API call
py <- plotly()
py$plotly(fig$data, kwargs=list(layout=fig$layout))
```

It could be an opportunity to rewrite the `plotly()` method (as in `py$plotly()`), making things explicit and less 'flexible' (dot-dot-dot argument). But this is not as urgent so I have only adapted to the new figure return (7f4435c).

Tests won't pass as of commit 6b2a8e8, I'm going to make extensive use of `sed` to update our tests accordingly.

/cc @chriddyp @etpinard @cpsievert @tdhock


cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/60482828 

 On TravisCI, commit d39157d5e65ebceea7ffb0284b4298e679564ad8 was successfully merged with d9ecaf43fa75eeef0142cef93361fdd1e936d227 (master) to create 4e02b73c95a648a454c05e450187600f0d896b5c. A visual testing table comparing d9ecaf43fa75eeef0142cef93361fdd1e936d227 with 4e02b73c95a648a454c05e450187600f0d896b5c can be found here:
 http://ropensci.github.io/plotly-test-table/tables/4e02b73c95a648a454c05e450187600f0d896b5c/index.html

cpsievert:
Assuming we want to go forward with the current design, these changes look good to me...

However, since this introduces some backwards-incompatible changes, I thought I'd mention again that [having `plotly()` return a list of functions is confusing and unnecessary](https://github.com/ropensci/plotly/issues/40#issuecomment-93822272). 

If `plotly()` was an [S3 generic](http://adv-r.had.co.nz/OO-essentials.html#s3), it would solve our [documentation problems](https://github.com/ropensci/plotly/issues/206) and make the interface more intuitive for R users. If we go that route, it'd be convenient to attach a class (say `'plotly_figure'`) to the return value of `gg2list()`. That way, `plotly()` could take either a ggplot object or the return value of `gg2list()` as an input.

```r
plotly <- function(p, username, key, ...)  UseMethod("plotly")
```

```r
plotly.plotly_figure <- function(p, username, key, ...) {
  if (missing(username)) username <- Sys.getenv("PLOTLY_USER")
  if (missing(key)) key <- Sys.getenv("PLOTLY_KEY")
  # make call to API and return response
  make_call(p, username, key, ...)
}
```

```r
plotly.default <- function(p, username, key, ...) {
  if (missing(username)) username <- Sys.getenv("PLOTLY_USER")
  if (missing(key)) key <- Sys.getenv("PLOTLY_KEY")
  # make call to API and return response
  make_call(gg2list(p), username, key, ...)
}
```

Of course, making this change would cause more serious backwards-incompatible changes, which will upset users, but making the necessary changes shouldn't be that hard. @mkcor @chriddyp if you're willing to make these changes, please let me know, and I'll start a pull request.

mkcor:
@cpsievert Thanks for your insight. I think this conversation belongs there https://github.com/ropensci/plotly/issues/40 so I'll reply there about your re-design suggestions. I understand it's a related issue as far as we're introducing backwards incompatibility. I admit, I'm not very concerned with backwards incompatibility here (maybe I should care a lot, just by principle) because I doubt that many projects around the world rely on our `gg2list()` function... And even if they do... well, I know, I'm totally betraying my identity here ("Scientists want to explore. Engineers want to build." see http://www.software.ac.uk/blog/2015-02-06-scientific-coding-and-software-engineering-whats-difference)!

mkcor:
@cldougl Once this is merged, `gg2list()` will return the expected figure object, so that there is no need for wrapping function `gg2fig()` (https://github.com/cldougl/plotly-shiny/blob/master/What_You_Need/plotlyGraphWidget.R#L18-L25).

chriddyp:
I'm not too worried about backwards incompatibility since this is a private function anyway. This is a good step forward for usability in Shiny and some other customizability, so :+1: !

@cpsievert , @mkcor , @13bzhang , @cldougl - let's mediate on the redesign of #40 this week and convene sometime next week with a more thorough discussion!

mkcor:
Thanks, @chriddyp !

Strictly speaking / technically, `gg2list` is not private, because you can call it readily once you've loaded the "plotly" package, and we've not using `@keywords internal` in its docstring or anything.
But I would say that, *culturally*, it is, indeed.

Okay, I'll sync up with `master` and update files `NEWS` and `DESCRIPTION`.

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/61206866 

 On TravisCI, commit aff299724bcef81fa7a8a9cb163e4254fa36906b was successfully merged with d9ecaf43fa75eeef0142cef93361fdd1e936d227 (master) to create 512b5b501820c1e22133657ac4472e5b98a72bcc. A visual testing table comparing d9ecaf43fa75eeef0142cef93361fdd1e936d227 with 512b5b501820c1e22133657ac4472e5b98a72bcc can be found here:
 http://ropensci.github.io/plotly-test-table/tables/512b5b501820c1e22133657ac4472e5b98a72bcc/index.html

cpsievert:
I agree that these "backwards-incompatible" changes aren't a huge deal; but nevertheless, I like to follow [the semantic versioning protocol](http://semver.org/), which would suggest to bump the version to 0.6.

Sorry I didn't mention that earlier. Again, not a big deal, but something to consider for the future.

mkcor:
@cpsievert Great call! You are right. I care about versioning properly. Submitting https://github.com/ropensci/plotly/pull/215

Issue #209, opened 2015-04-28
Tests are run twice


cpsievert:
The current approach to testing is to run tests via `R CMD check` and then run tests *again* to build the visual testing table. We could reduce build time by doing something like (inside `save_outputs()`?):

```r
if (Sys.getenv("USER") == "travis") {
  # build table while running tests
} else {
  # just run the tests 
}
```

Issue #204, opened 2015-04-15
Unable to set credentials, 'No such file or directory'


arnoud999:
When I try to set my credentials/API key, I get the following error message:

    [1] "Your credentials file:"
    Error in if (file.info(CREDENTIALS_FILE)$size) { : 
      argument is not interpretable as logical
    In addition: Warning message:
    In file.create(abspath) :
      cannot create file '//studfiles.campus.uvt.nl/files/home/home09/u1236118/.plotly/.credentials', reason 'No such file or directory'

I've tried creating the .plotly folder and the .credentials file myself, and I tried running R through a command window as administrator, but in both cases I still saw the same error message.

theengineear:
@arnoud999 , are you on a personal computer? The credentials file tries to get created in `~/.plotly/`. That error seems to be saying `~/.plotly` doesn't exist. As a test, can you open up a terminal and do the following?

```
cd ~/.plotly
```

You've probably got a permissions problem, but I'm sure we can figure out a way to get you up and running.

The *simplest* thing to get going straight-away is to just sign in.

`py <- plotly("your_username", "your_api_key")`

In general, this isn't a great solution as you'll be hard-coding your credentials and you'd have to do it for *every* script.

arnoud999:
Thanks for your quick response!

`cd ~/.plotly` (or `cd ~\.plotly`) doesn't work, probably because I'm running Windows? 

It indeed seems to be a permission problem, I'm on a work computer and my user folder is some kind of network drive. Is it possible to change the location of `.plotly`?

theengineear:
@arnoud999 , specifying the location of the `.plotly` folder could be nice, I think that would be pretty easy through an environment variable. We won't have this solution for you immediately, but I could definitely see the utility in this, so I'll see what I can do.

Until then, yes, you'll need to sort out the permission block. I had thought that our R library handled this though, @mkcor , thoughts?

mkcor:
@theengineear You are right, I remember troubleshooting Windows users in the past...  But this may be yet another thing...  It could be a thing to let users specify/overwrite environment variable `PLOTLY_DIR`...

I think Windows does not understand `~` (which is why our R code needs the `path.expand("~")` by the way). @arnoud999 It looks like the 'home' on this workstation is not quite the home the way we usually assume.

Can you try and run, say:
```r
py <- plotly("rAPI", "yu680v5eii")
```
Please?

arnoud999:
Unfortunately, I get the same error:

[...]

PR #217, opened 2015-05-07
API redesign

cpsievert:
I couldn’t help myself and took a stab at the redesign discussed in #40.

Some of the big changes include:
  1. No more `list$function()` interface. The new interface should be more intuitive to R users and resolves documentation issues (see #206). See the example section in `help(plotly)` for an overview of the new interface.
  2. Credentials/configuration are managed via environment variables so file permissions shouldn’t be an issue anymore (see #204). Users *can* still pass their username and API key to `plotly()` (and they will be automatically stored as environment variables). Also, the `signup()` function will (by default) try to store these variables in `~/.Rprofile` so they will be loaded upon startup.
  3. Uses **httr** instead of **RCurl** to communicate with the plotly API. As a result, the source code is much simpler and provides some useful utilities for turning HTTP errors into R errors.
  3. Leverages the `knitr::knit_print()` S3 generic to automatically embed plotly iframes in R Markdown files. See `test-plotly-knitr.R` for a simple example.
  4. IPython users will have to use `embed_notebook()` to embed plotly frames in IPython (I haven't tested this yet, it'd be great if others could test this functionality).

Since this is such a drastic change to API, merging this would require bumping to 1.0.0. I’d be happy to work on shiny bindings as well (I’ve done this for several other R packages), but that might be appropriate for another pull request.


cpsievert:
Since there are backwards incompatible changes, some changes will have to be made to the build script in plotly-test-table. I'll be experimenting with this over on [my fork](https://github.com/cpsievert/plotly-test-table) and might take this as opportunity to fix #209 and simplify the build process.

mkcor:
I must say that development has been strongly guided and influenced by the Python library... For instance, `tools.R` was a direct translation of https://github.com/plotly/python-api/blob/master/plotly/tools.py! Likewise, resorting to the `signup` function has been deprecated.

cpsievert:
I'd be disappointed if this PR falls on deaf ears because it doesn't mimic the Python implementation...it's considerably simpler for both users and developers.

Why was the `signup()` function deprecated?

Just FYI, it looks like the os module can handle environment variables as well -- http://nbviewer.ipython.org/github/lightning-viz/lightning-example-notebooks/blob/master/images/images.ipynb

theengineear:
@cpsievert , the `signup()` function was deprecated because we were seeing code go public that reveals private information (api keys). The easiest way we came up with to fix this pattern was to try and nudge folks into using the credentials/config files.

As for environment variables. That seems like a totally reasonable approach. If we do it for *one* library, the others should all support it though as the idea is transparent cross-collaboration between languages.

theengineear:
Could you split out the credentials stuff into a different PR so the rest doesn't get held up?

cpsievert:
> the `signup()` function was deprecated because we were seeing code go public that reveals private information (api keys).

At least for the R package, it'd be easy to not relay the API message to avoid that issue. I like the `signup()` function, because to get started, all we do is:

```r
library(plotly)
signup("me", "me@gmail.com")
q <- qplot(1:10)
plotly(q)
```

> Could you split out the credentials stuff into a different PR so the rest doesn't get held up?

That would introduce more backward incompatibility than we really need. Also, considering the integration with [plotly-test-table](https://github.com/ropensci/plotly-test-table)), it will go much more smoothly if we do everything at once. Give me a couple days and I'll have a new build process for the test table which will be self-contained (within this package), more efficient, and easier to maintain.

cpsievert:
> If we do it for one library, the others should all support it though as the idea is transparent cross-collaboration between languages.

I can certainly see the benefit of sharing credentials between languages. If you'd like to keep the current approach of writing credentials to disk [like this](https://github.com/plotly/python-api/blob/master/plotly/tools.py#L42-L68), just let me know, and I can add some support for reading/writing to those files (and fall back on environment variables if we don't have file permissions).

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/62281818 

 On TravisCI, commit a321dc3 was successfully merged with 789e1ec (master) to create e13078b. A visual testing table comparing 789e1ec with e13078b can be found here:
 http://cpsievert.github.io/plotly-test-table/R/e13078b/index.html

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/62286566 

 On TravisCI, commit 5d5499b was successfully merged with 789e1ec (master) to create 6916f26. A visual testing table comparing 789e1ec with 6916f26 can be found here:
 http://cpsievert.github.io/plotly-test-table/R/6916f26/index.html

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/62334253 

 On TravisCI, commit c0f1ca9 was successfully merged with 789e1ec (master) to create 3aaf483. A visual testing table comparing 789e1ec with 3aaf483 can be found here:
 http://cpsievert.github.io/plotly-test-table/R/3aaf483/index.html

bpostlethwaite:
We are in the process of rolling out v2 of our API https://api.plot.ly/v2/ (60% complete, 100% complete in early June) and I would suggest any rewrite utilizing the new APIs as the existing calls will begin to return deprecation warnings. Things like exposing `kwargs` as an API parameter are, thankfully, a thing of the past.

cpsievert:
That's great to know, and I plan on waiting for v2 before merging this, thanks @bpostlethwaite!

The biggest thing for me is having the equivalent of the "clientresp" resource in the old API. It looks as though that will go under here https://api.plot.ly/v2/plots, but it's not quite ready yet?

bpostlethwaite:
@cpsievert yes, that is correct. `ClientResp`'s replacement `plots` is the most work and will likely not land until early June. 

@theengineear and @BRONSOLO are rolling it out and can let you know, and take questions, once it is complete.

Issue #218, opened 2015-05-13
Equal scaling of axes not working


13bzhang:
This is one of the unsupported features from the cookbook, see Issue #161.

There's a problem with setting the scaling of the axes to an equal ratio, with one visual unit being representing the same numeric unit on both axes. There's also a problem when I set them to ratios other than 1:1.

```
# Data where x ranges from 0-10, y ranges from 0-30
set.seed(202)
dat <- data.frame(
    xval = runif(40,0,10),
    yval = runif(40,0,30)
)
sp <- ggplot(dat, aes(xval, yval)) + geom_point()

# Force equal scaling
sp + coord_fixed()
```

Cookbook:
![1-1_scaling](https://cloud.githubusercontent.com/assets/6131842/7619028/0be0c3be-f984-11e4-9f4b-435e75a6d29b.png)

Plotly:
![1-1_scaling_plotly](https://cloud.githubusercontent.com/assets/6131842/7619052/37f48904-f984-11e4-9cde-d961dc05f08a.png)

```
# Equal scaling, with the total x axis the same length as the total y axis
sp + coord_fixed(ratio=1/3)
```
Cookbook:
![1-3_scaling](https://cloud.githubusercontent.com/assets/6131842/7619084/58f61910-f984-11e4-9984-ec331c0f88da.png)

Plotly:
![1-3_scaling_plotly](https://cloud.githubusercontent.com/assets/6131842/7619101/77c349e4-f984-11e4-9c69-a6dec554cf10.png)

cc @cpsievert @chriddyp @mkcor

13bzhang:
Here's a solution I came up with:

https://github.com/ropensci/plotly/blob/baobao-equal_axes/R/ggplotly.R#L771-785
```
  # fixed coordinates: if the coordinates ratio is not NULL, then
  # we make the size of the plot according to the specified ratio
  # note: we set the biggest dimension to 600
  if (!is.null(p$coordinates$ratio)){
    x.range <- range(built[[2]]$ranges[[1]]$x.major_source, na.rm = TRUE)
    y.range <- range(built[[2]]$ranges[[1]]$y.major_source, na.rm = TRUE)
    y.x.ratio <- (y.range[2] - y.range[1])/(x.range[2] - x.range[1])
    if (y.x.ratio > 1){
      layout$height <- 600
      layout$width <- layout$height*(1/p$coordinates$ratio)*(1/y.x.ratio) 
    } else{
      layout$width <- 600
[...]

PR #222, opened 2015-05-15
Equal scaling of axes

13bzhang:
This pull request is a solution to fix Issue #218 "Equal scaling of axes not working". I fixed it so the width and height of Plotly plots have the ratios specified by ggplot2 user if they decide to specify coordinate ratios.


mkcor:
Aww a `.Rhistory` file should never end up in a PR...  Please use
```bash
$ git add <file>
$ git commit -m "commit message"
```
(suspecting you used `git commit -am ...`?) but a `.Rhistory` file should never be tracked by your VCS in the first place.  I'm not sure it's enough to remove it with an additional commit, because I wouldn't want it anywhere in our commit history.

@13bzhang Can you please submit a new PR and I'll close this one? Thank you so much!

mkcor:
Please add spacing around `*` operators and... I think the dotted notation is not our friend for simple variable names (the R culture is changing too: http://kbroman.org/hipsteR/).

PR #228, opened 2015-06-05
Fix element-wise access in for loop

mkcor:
Ran into this bug by chance, while working on content.

It might be `g$aes[a]` instead of `g$aes[a.name]`... I should come up with an example where that makes a difference and add it as a test. But clearly it cannot be the vector we're looping over (`aes.used`).

/cc @cpsievert @13bzhang @chriddyp


mkcor:
The test is about converting this ggplot:

segment-multiple

See [source code](https://plot.ly/~marianne2/1718).

At the moment (before this fix), the conversion (or making of the figure object) simply fails. Error being
```
Error in .subset2(x, i, exact = exact) : subscript out of bounds
```

With this fix, the figure object (obtained by calling `gg2list()` onto this ggplot) is exactly what you would expect. But the final plotly plot is missing the third trace! See https://plot.ly/~marianne2/1718

If you do a `get_figure()`, you can see that there are four traces as expected, but the last two (the `segment` ones, of interest) are duplicates (of the second segment, around `x = 2`).

So we have an issue, here, I'm confused, since it looks like we're sending the right figure object.
/cc @cpsievert @13bzhang @chriddyp @corinne-riddell

So this fix is making things a lot better but they're still a little broken...

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/95922020 

 Detected 1 differences -> 

 http://cpsievert.github.io/plotly-test-table/53fea51/scatterplots-jitter

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/96070490 

 Detected 1 differences -> 

 http://cpsievert.github.io/plotly-test-table/a0ad70f/scatterplots-jitter

PR #243, opened 2015-07-31
add .embed to iframe

chriddyp:
otherwise plotly will detect that the graph is in an iframe, and redirect to `.embed` anyway. so, this should make plotly embeds load ~2 seconds faster :)

cc @cpsievert


cpsievert:
cool, thanks!

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/73571195 



On TravisCI, commit 095c814 was successfully merged with 4b4f774 (master) to create 085e8f3. A visual testing table comparing 4b4f774 with 085e8f3 can be found here -> 

 http://cpsievert.github.io/plotly-test-table/R/085e8f3/

Detected 0 differences

Issue #247, opened 2015-08-04
`ERROR: dependency ‘viridis’ is not available for package ‘plotly’`


chriddyp:
hey @cpsievert  - not sure about the best practices for dependencies, but is it possible to include `viridis` as a requirement that gets automatically installed? seems like folks have to `install.packages('viridis’)` before the install `plotly`

cpsievert:
Ah, it's [listed under IMPORTS](https://github.com/ropensci/plotly/blob/5cdd985a0d5eb1479cf972fc1712c4125025f8e2/DESCRIPTION#L33), but it must not install by default since it didn't appear in the NAMESPACE as well. https://github.com/ropensci/plotly/commit/ce587aa0a2fc6ddc9c4abfab7283ea1666ff3df8 should fix it

chriddyp:
I'm still getting this issue with a fresh install

cpsievert:
Is the problem with installing viridis, or the problem reported in #253?

13bzhang:
I have been getting this error message when trying to install `viridis` by `install.packages("viridis")`. I can install it fine via GitHub. But the package is on CRAN, so...

```r
Warning in install.packages :
  package ‘viridis’ is not available (as a binary package for R version 3.1.3)
```
I have R version 3.1.3 on my OS X.

chriddyp:
The problem being that the library isn't installed automatically as a dependency:

```
> devtools::install_github("ropensci/plotly")
Downloading github repo ropensci/plotly@master
Installing plotly
Skipping 1 packages not available: viridis
Installing 11 packages: colorspace, dichromat, ggplot2, gtable, labeling, munsell, plyr, proto, RColorBrewer, reshape2, scales
'/Library/Frameworks/R.framework/Resources/bin/R' --no-site-file --no-environ  \
  --no-save --no-restore CMD INSTALL  \
  '/private/var/folders/d4/c4jpnb754pj3cfj5pv8y9m080000gn/T/Rtmp5ZPPeo/devtools5dcc767c747/ropensci-plotly-a97299c'  \
  --library='/Library/Frameworks/R.framework/Versions/3.1/Resources/library'  \
  --install-tests 

ERROR: dependency ‘viridis’ is not available for package ‘plotly’
* removing ‘/Library/Frameworks/R.framework/Versions/3.1/Resources/library/plotly’
Error: Command failed (1)
```

13bzhang:
[...]

PR #257, opened 2015-08-11
Loading dependency "viridis"

13bzhang:
This is a temporary fix for loading dependency "viridis." I created a zzz.R file with a onLoad funciton that checks if `viridis` is installed. If not installed, I tell it to install it from `type = "source"`.


chriddyp:
nice! looks good to me, @cpsievert ?

cpsievert:
no, you shouldn't install packages for users, see [my comment](https://github.com/ropensci/plotly/issues/247#issuecomment-130055710) in #247

PR #279, opened 2015-09-04
Add support for images endpoint

cpsievert:


cpsievert:
@chriddyp `plotly_IMAGE()` seems to work for me locally now. Since I'm mostly adding this to reduce overhead in our testing infrastructure, I'll go ahead and make those changes to testing in this pull request (unless you'd like it separated out in another PR).

cpsievert:
@chriddyp this is ready for merging. It shouldn't change any existing functionality, it essentially just adds the `plotly_IMAGES()` function which should be useful for testing.

If we merge now, then I can assume master's HEAD has `plotly_IMAGES()`, which will make it easier to modify the test table build script to use the images endpoint.

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/80260522 



On TravisCI, commit de951de was successfully merged with b0532b9 (master) to create d34e75f. A visual testing table comparing b0532b9 with d34e75f can be found here -> 

 http://cpsievert.github.io/plotly-test-table/R/d34e75f/

Detected 0 differences

PR #334, opened 2015-12-10
Fix #333

### LINKED ISSUE #333: set the R config variable `plotly_domain` inside JS as `window.PLOTLY_ENV.BASE_URL`
In `plotly.js`, we use `window.PLOTLY_ENV.BASE_URL` (https://github.com/plotly/plotly.js/blob/master/src/plots/plots.js#L370) to determine which cloud to send the graph to. If the user set `plotly_domain` inside their R `ENV`, we need to set that in the JS `window.PLOTLY_ENV.BASE_URL` so that charts don't get sent outside the network.

cpsievert:


inst/htmlwidgets/plotly.js
```diff
@@ -11,6 +11,9 @@ HTMLWidgets.widget({
   },  
   
   renderValue: function(el, x, instance) {
+    // make sure plots don't get created outside the network
+    window.PLOTLYENV.BASE_URL = x.base_url;
```
chriddyp: In case it doesn't exist:

```javascript
window.PLOTLYENV = window.PLOTLYENV || {};
window.PLOTLYENV.BASE_URL = x.base_url;
```

cpsievert: Thanks!

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/96076387 

 Detected 1 differences -> 

 http://cpsievert.github.io/plotly-test-table/05e169f/scatterplots-jitter

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/96073604 

 Detected 1 differences -> 

 http://cpsievert.github.io/plotly-test-table/39c2343/scatterplots-jitter

Issue #375, opened 2015-12-26
LaTeX Typesetting in R doesn't work


statist-bhfz:
I can't reproduce an example from https://plot.ly/r/LaTeX/ with latest versions of R, RStudio and all packages on Windows (including development version of plotly). It looks like something wrong with MathJax support in the current version of package.
I have provided some screenshots in my question on Russian StackOverflow: http://ru.stackoverflow.com/questions/478133/latex-%D0%B2-plotly-r/478168#478168

artemklevtsov:
Should be fixed if add mathjax library to the package and htmlwidgets config.

cpsievert:
Mathjax is too big to bundle with the R package, so we're looking at a situation similar to #302, where we want to conditionally download/include relevant stuff under https://github.com/plotly/plotly.js/tree/master/dist before rendering

The thing is, with the current interface, there is no way to detect whether MathJax is needed. Maybe we provide a simple function, say

```r
tex <- function(x) {
  structure(x, class = "LaTeX")
}
```

which could then be used essentially as a flag to include MathJax:

```r
plot_ly(x = c(1, 2, 3, 4), y = c(1, 4, 9, 16),
        name = tex("\\alpha_{1c}"))
```

Any thoughts @chriddyp @cldougl @etpinard?

cpsievert:
@timelyportfolio a while ago I tried to resolve this in #411. I'm sure you can come up with a better approach :)

timelyportfolio:
Would it be reasonable to expect RStudio and `rmarkdown` availability?  If so, perhaps we could leverage the `rmarkdown:::pandoc_mathjax_local_path()` function to use that or just educate users on [`mathjax: local`](http://rmarkdown.rstudio.com/html_document_format.html#mathjax_equations) option in `html_document`.  Then we wouldn't have to worry about doing all that RStudio does (see [lines](https://github.com/rstudio/rmarkdown/blob/master/tools/prune-mathjax.sh)).

So, this works for me

```
---
title: "MathJax Test"
runtime: shiny
output:
  html_document:
    mathjax: local
---

$$ 
   \frac{d}{dx}\left( \int_{0}^{x} f(u)\,du\right)=f(x).
$$

    ```{r, echo=FALSE} 
    sliderInput("x", label = "Value:",  min = 1, max = 10, value = 5)

    renderText({
       input$x
    })
    ```
```

To confirm local, here is a screenshot of the network request.


[...]

Issue #356, opened 2015-12-16
Choropleth Maps do not appear to display in RStudio viewer


pssguy:
or RMarkdown

cpsievert:
Currently you need an internet connection to view maps. In #302, we're toying with the idea of including the assets required to make maps with the R package, so they can be created/viewed offline

PR #302, opened 2015-11-17
add geo assets so maps render without internet connection

cpsievert:

PR #411, opened 2016-01-19
Copy extras over from plotly.js, if necessary

cpsievert:
This addresses #375 and #356 by leveraging a new environment variable which should point to the plotly.js source. In other words, you can render MathJax and/or maps without an internet connection by doing something like:

``` shell
git clone https://github.com/plotly/plotly.js.git
cd plotly.js
export plotly_jsdir=`pwd`
echo "Sys.setenv('plotly_jsdir', '$plotly_jsdir')" >> ~/.Rprofile
```


cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/103455036 

 Detected 3 differences -> 

 http://cpsievert.github.io/plotly-test-table/21067bc/jitter-basic
http://cpsievert.github.io/plotly-test-table/21067bc/scatterplots-geom_jitter
http://cpsievert.github.io/plotly-test-table/21067bc/scatterplots-jitter

cpsievert:
MathJax isn't quite working yet as you can see here -- http://cpsievert.github.io/widgetHub/

cpsievert:
TODO: why does this warning occur?

```r
> plot_ly()
Warning message:
In file(con, "r") :
  file("") only supports open = "w+" and open = "w+b": using the former
```

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/103694301 

 Detected 3 differences -> 

 http://cpsievert.github.io/plotly-test-table/dd886f8/jitter-basic
http://cpsievert.github.io/plotly-test-table/dd886f8/scatterplots-geom_jitter
http://cpsievert.github.io/plotly-test-table/dd886f8/scatterplots-jitter

cpsievert:
Also, [injecting `<script>` to load mathjax via CDN isn't working](https://github.com/ropensci/plotly/blob/a7d61b261aaaa5126538ef2df16c054b979cb2b1/inst/htmlwidgets/plotly.js#L22-L28), and I don't think their a nice way to declare a non-local dependency with htmltools/htmlwidgets （╯°□°）╯ ┻━┻

timelyportfolio:
I am not sure of precedent here with other packages, so I will look for other examples of similar functionality.  There is this script in `rmarkdown` to prune `MathJax` to a smaller size.  Here are a couple of poorly named functions to download the additional assets from Plotly for maps and MathJax to start the discussion/iteration.  Then we will need some add dependency functions to add the assets to the Plotly htmlwidget.  We have the option to do either local or web assets in these added dependencies.

```
plotly_download_map_dep <- function(){
  # find location of htmlwidgets assets in plotly package
  plotly_dir <- system.file("htmlwidgets/lib",package="plotly")
  
  # check to see if directory already exists
  exists_map_dir <- 'plotlymaps' %in% basename(plotly_dir)
  
  maps_dir <- file.path(plotly_dir,"plotlymaps")
  
  if(!exists_map_dir){
    dir.create(maps_dir)
  }
  download.file(
    url = "https://cdn.rawgit.com/plotly/plotly.js/master/dist/plotly-geo-assets.js",
    destfile = file.path(maps_dir,"plotly-geo-assets.js")
  )
  return(file.path(maps_dir,"plotly-geo-assets.js"))
}
```

cpsievert:
Thanks @timelyportfolio. I think I'd prefer to grab assets from the relevant plotly.js release via GitHub's API, sort of like the `inst/plotlyjs.R` script

timelyportfolio:
@cpsievert, can you explain why?  This would require downloading the whole repo for a couple of files.

cpsievert:
True, I'm mainly concerned about versioning of the assets. As @chriddyp mentions [here](https://github.com/ropensci/plotly/pull/302#issuecomment-157763691), maybe we could grab the assets from our CDN (and hopefully those are versioned)?

timelyportfolio:
I was looking for the cdn location.  Thanks for the pointer.  Probably better to use that.

fabiangehring:
I didn't follow the discussion in the different issues about making geo plots available...

But if size is an issue, why not offering the offline content in another package like "plotlyOffline", comparable to "shinythemes" and "shiny". Plotly could then stay the same and gaining an "asset"-config with values "auto", "offline" or "online". Auto could check if the required assets are available and use them if yes and download otherwise. Different asset versions could be managed like normal package version dependencies. 

I think this would keep plotly small but available "offline" if necessary. And As far as I can tell the implementation shouldn't be too difficult?

danielsjf:
What is the current status of this development? I like the idea of @fabiangehring as this practice is indeed used for many other R packages. The usage of online VS offline could be handled with an argument in the function or an option.

Specifically for the use case of #302, it is annoying that the maps are not shown in Rstudio. This has been an issue since late 2015.

cpsievert:
@danielsjf at this point, I would like have to rethink this PR and probably create a helper package, say `plotlyDependencies`, which would bundle and automatically infer which (minimal amount) of [plotly assets](https://github.com/plotly/plotly.js/tree/master/dist) are needed for rendering.

Unfortunately, at the moment, we don't fund have the time/funding. [Please get in touch](https://plot.ly/products/consulting-and-oem/) if you're interesting it supporting

cpsievert:
Closing in favor of #1283

eddy-ojb:
@cpsievert @chriddyp Did this get solved? I am running plotly Dash on an offline server and struggling to get the geo assets to load for maps. Specifically, Plotly is looking for assets like https://cdn.plot.ly/usa_110m.json. Is there a new method to render this locally?

PR #442, opened 2016-01-29
modify plotly_domain env var when using API v2, fixes #441

### LINKED ISSUE #441: plotly_POST sharing="secret" not working
When I publish a plot (created with ggplot2 and ggplotly) to my company self hosted plotly using
plotly_POST(p, filename="a_plot", sharing="secret")

I then have to go manually to my plotly file, click on the share link, and enable the secret sharing which I think is not what should happen.

I also tried to use "hidden" as the error message doesn't agree with the documentation.

cpsievert:


cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/105793432 

 Detected 3 differences -> 

 http://cpsievert.github.io/plotly-test-table/d77b7f3/jitter-basic
http://cpsievert.github.io/plotly-test-table/d77b7f3/scatterplots-geom_jitter
http://cpsievert.github.io/plotly-test-table/d77b7f3/scatterplots-jitter

PR #450, opened 2016-02-03
modified example to load ggplot2movies

martinjhnhadley:
movies is no longer contained within ggplot2 https://cran.r-project.org/web/packages/ggplot2movies/index.html


cpsievert:
Thanks! A couple of tests fail just because of a credentials issue...

Issue #418, opened 2016-01-21
Using `group` + `color` produces incorrect plot


alyst:
`plotly` incorrectly draws lines `scatter` (and `scatter3d`) plots.

Consider the following minimal example:
```R
df = data.frame(x=c(1,2,3,4,5,6),
           y=c(1,1,2,2,3,3),
           t=c("A","A","B","B","A","A"))

plot_ly(df, type="scatter", x=x, y=y, color=t, group=y, mode="lines")
````

The output is:
![image](https://cloud.githubusercontent.com/assets/348591/12484986/08bd8212-c05d-11e5-9d3c-61b560a47bef.png)

In the correct plots only the horizontal segments colored according to `t` column are expected; also no `1`,`2`,`3` entries in the legend, `A` and `B` sorted either alphabetically or by their appearance in the data frame.

**Comment by cpsievert** [comment]
I have a note of this here as well #381

I hope to get to this soon...thanks for the minimal example!

**Comment by alyst** [comment]
@cpsievert Yes, it's the same as #381 + an additional issue of drawing "parasitic line" `(2,1) - (5,3)`.

**Comment by cpsievert** [comment]
Closed via d2bf2a1eb56d824a091e2cc9afea09a5968dd050. This is now the official way to do this:

```r
df = data.frame(x=c(1,2,3,4,5,6),
           y=c(1,1,2,2,3,3),
           t=c("A","A","B","B","A","A"))

df %>% group_by(y) %>% plot_ly(x = ~x, y = ~y, color = ~t) %>% add_lines()
```

PR #471, opened 2016-02-23
RFC: fix traces subdivision by multiple properties

### LINKED ISSUE #381: group should not create multiple traces
The `group` argument in `plot_ly()` should probably be [translated the same way as ggplot2's group aesthetic](https://github.com/ropensci/plotly/wiki/ggplotly-design,-internals,-technical-stuff).

Among other things, this would make it easier to create candlestick charts requested in #351

alyst:
This is my version of the #418 fix:
- add `indices` property to the trace to reference the indices of
  the data elements associated with the trace
- replace `tracify()` by `subdivide_traces()` that correctly subdivides
  the current set of traces based on the given data property

Note that while the traces are assigned the correct visual properties, the PR does not fix #381 -- individual `group` traces are still being created and displayed in the legend.

Comments, especially regarding the ways to fix #381 are most welcome.

### DIFF
--- a/R/plotly.R
+++ b/R/plotly.R
@@ -292,8 +292,7 @@ plotly_build.plotly_hash <- function(l = last_plot()) {
   # this is ugly, but I think it is necessary, since we don't know how many 
   # traces we have until we evaluate args and call traceify() (or similar)
   x <- list()
-  for (i in seq_along(l$data)) {
-    d <- l$data[[i]]
+  x$data <- unlist(lapply(l$data, function(d) {
     if (should_eval(d)) {
       dat <- do_eval(d)
       # start processing specially named arguments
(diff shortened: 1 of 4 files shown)


cpsievert:
Wow! At first glance this looks great and attacks a big problem :beers: 

I'll need to a few days to mull this over. In the meantime, could you add some tests to ensure this new behavior works as intended?

Also, don't worry about the current test failures... I need to find a way to skip those on external pull requests.

R/plotly.R
```diff
@@ -254,48 +254,64 @@ plotly_build <- function(l = last_plot()) {
...
+        traces <- subdivide_traces(traces, dat, "legenditem", function(sub_trace, lvl) {
+          sub_trace$legendgroup <- legenditems[[lvl]]
+          sub_trace$name <- legenditems[[lvl]]     # override the trace name by legendgroup
+          sub_trace$showlegend <- showlegend[[lvl]] # put each group to legend once
+          if (showlegend[[lvl]]) {
+            showlegend[lvl] <<- FALSE
+          }
+          sub_trace
+        })
+      }
```
cpsievert: @alyst I'm not quite sure what this if statement is doing. Should this say `legendgroup` instead of `legenditem`?

R/plotly.R
```diff
@@ -254,48 +254,64 @@ plotly_build <- function(l = last_plot()) {
...
   # traces we have until we evaluate args and call traceify() (or similar)
   x <- list()
-  for (i in seq_along(l$data)) {
-    d <- l$data[[i]]
+  x$data <- unlist(lapply(l$data, function(d) {
     if (should_eval(d)) {
       dat <- do_eval(d)
+      # put everything into a single trace
+      trace <- dat
+      trace[["indices"]] <- seq_along(trace[["x"]]) # indices of the original data elements used in the trace FIXME properly define data length
```
cpsievert: @alyst what happens if `trace[["x"]]` is `NULL`?

R/plotly.R
```diff
@@ -254,48 +254,64 @@ plotly_build <- function(l = last_plot()) {
   # this is ugly, but I think it is necessary, since we don't know how many 
   # traces we have until we evaluate args and call traceify() (or similar)
   x <- list()
-  for (i in seq_along(l$data)) {
-    d <- l$data[[i]]
+  x$data <- unlist(lapply(l$data, function(d) {
```
cpsievert: Is the `unlist()` necessary here?

R/plotly.R
alyst: Nothing good. :) As I indicate here, I just need some reliable way to get the trace length. Do you know a better alternative? (Maybe it would be nice to have `trace_length(trace)` utility function)

R/plotly.R
alyst: @cpsievert Yes, the function inside `lapply()` returns a list of traces for each `l$data` element. `unlist()` just combines these lists into one.

R/plotly.R
cpsievert: The length should probably come from variable you're splitting on...

R/plotly.R
alyst: @cpsievert It tries to solve the problem of having an item in the legend for each trace (which is not what you want if e.g. you are doing something similar to `geom_segment()`). `legenditem` specifies the name of the _single_ element in the legend that all the traces in the same `legenditem`-group would be referring to (unless they are further subdivided by something else).
To implement this, only the first trace of the group are shown in the legend, but all the other traces still share the same `legendgroup`.
Although it utilizes `legendgroup` mechanism, I thought it's better to call the property `legenditem`, because from the user POV it doesn't do any grouping of legend elements, it just creates one item for the group of traces.

R/plotly.R
alyst: @cpsievert That would make sense, but here we try to index every point in the yet-unprocessed trace before any particular splitting is conceived.

R/plotly.R
cpsievert: I should have mentioned this a long time ago in response to [your comment](https://github.com/ropensci/plotly/issues/381#issuecomment-171343812):

Instead of creating a trace for each group and managing legend entries via `legendgroup`, I'd prefer to take the same approach as `ggplotly()` where `group` doesn't create multiple traces, but groups are separated by `NA`s. For a simple example, `plot_ly(x = c(1:10), y = c(1:10), group = rep(1:2, each = 5))` should translate to `plot_ly(x = c(1:5, NA, 6:10), y = c(1:5, NA, 6:10))`.

We use `plotly:::group2NA()` to handle this `ggplotly()`. Note that [this line](https://github.com/ropensci/plotly/blob/d8d8d65d61d3fbfe5a87818d1133a72bfd3a5054/R/layers2traces.R#L592) checks to see if we are drawing a polygon (in plotlyjs, a polygon is a scatter trace with mode of lines). In that case, we need to "retrace" the first point in each group to effectively close each polygon. There isn't any nice, high-level way to say "draw a polygon" using `plot_ly()`, so for now, I think we could just slip in `NA`s without worrying about retracing points. Does that make sense @alyst?

alyst:
@cpsievert Oh, I wasn't aware about the NA trick. Definitely it makes things a lot easier! I'll try to update the PR to make use of it (actually, it would also be easier to implement (sub)trace grouping using `dplyr::group_by()` & co, because current loop-based implementation is quite slow for medium to large-sized datasets).
However, in your example `group` serves for defining the subtraces and disappears during the postprocessing. What if somebody wants to use `group` to define the trace name? Maybe we can leave the `group` behaviour as is and introduce the `subgroup`(`subtrace`?) property that would be processed exactly as you described?

cpsievert:
> in your example group serves for defining the subtraces and disappears during the postprocessing.

What do you mean by subtraces? My example is really meant to show how the group argument _should_ work (i.e. it _should_ create just one trace,, but separated by missing values).

> What if somebody wants to use group to define the trace name?

I'm not sure I see the use case. Can you provide an example? In my example, since it _should_ create just one trace, you could name it like this: `plot_ly(x = c(1:10), y = c(1:10), group = rep(1:2, each = 5), name = "myTrace")`

> Maybe we can leave the group behaviour as is and introduce the subgroup(subtrace?) property that would be processed exactly as you described?

I don't think so. `group` should not create new traces. It should only control `NA` placement _after_ we've "traceified" for `color`/`shape`/etc.

alyst:
Ok, I see: the `color` and `shape` can tracify (i.e. introduce new traces), but the `group` can not. I had an idea that maybe `group` could also tracify, then it would have been possible to control the trace names via the `group` values. The use case is when both colors and shapes are mapped essentially to the same data column, then the autogenerated trace names in the legend would look strange, and it would be nice to use `group` to override it.
Of course, ATM one can do `add_trace(name=...)` for each trace, but that's rather tedious. Maybe we can allow binding `name` to a column and support tracifying by `name`?

alyst:
@cpsievert Based on the comments, I've updated (redesigned) the PR. It passes all the tests except those where interaction with the plotly server is required.

1. Tracification is now using `dplyr::group_by()` (hence `plyr` dependency (I have not detected where it is used currently) was replaced by `dplyr`). That is faster than looping through the unique levels of the data vector and it's done only once for all data vectors that affect tracification (e.g. `symbol`, `color` etc).
2. Tracification is done in 2 phases: first, the distinct traces (i.e. the indices of `dat` data vectors elements) are identified, then `dat` is split into traces by `generate_traces()` function accordingly.
3. The adjustment of trace visuals that was previously done by `colorize()`/`symbolize()` functions has been generalized into the "brushes" concept: `colorize()` is replaced by `color_brush()` function that  returns a "brush" -- the function that has to be applied to a trace to adjust its color. `generate_traces()` gets a list of the brushes to be applied to each trace. It should allow easier extension of the tracification logic.
4. As suggested, `group` property no longer generates the new traces (so 2 tests are disabled). Basically, it does nothing unless we are in "scatter" "lines" mode, in which case it inserts NA into the data at each group boundary to break the polylines. Since it utilizes `dplyr::group_by()`, it should be noticeably faster than the trace-based approach. It should be also faster on the client side.
5. I've kept the "legenditem" property that I proposed in the previous version of the patch. It would allow to group the traces with different visual properties under a single legend item. Should be useful when there are too many different traces to show in the legend. Another alternative would be to enable tracification for the `name`/`legendgroup`/`showlegend` properties.

cpsievert:
This problem was fixed by #628. Thanks for the fruitful discussion @alyst

Issue #463, opened 2016-02-13
Information not shown when hovering over data points in an ioslides presentation with multiple plots 


drwagner:
Hi

I'm trying to use Plotly with ioslides presentations and ran into the problem that hovering over the plot didn't show information for data points. This bug is somehow related to using multiple plots.

In the following example, I use the same plot on two slides. Hovering does not work with the first plot but it works with the second plot.

RMarkdown Document:

<pre><code>
---
title: "Plotly Bug"
output: ioslides_presentation
runtime: shiny
---

```{r, include=FALSE}
library(plotly)
```

## First Plot

```{r, echo=FALSE}
plot_ly(economics, x = date, y = uempmed, mode = "markers")
```

## Second Plot

```{r, echo=FALSE}
plot_ly(economics, x = date, y = uempmed, mode = "markers")
```
</code></pre>

Output of  devtools::session_info() 

<pre><code>
Session info ----------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.2.3 (2015-12-10)
 system   x86_64, darwin14.5.0        
 ui       RStudio (0.99.878)          
 language (EN)                        
 collate  de_DE.UTF-8                 
 tz       Europe/Berlin               
 date     2016-02-13                  

Packages --------------------------------------------------------------------------------------------------------
 package     * version date       source                            
 base64enc     0.1-3   2015-07-28 CRAN (R 3.2.0)                    
 colorspace    1.2-6   2015-03-11 CRAN (R 3.2.0)                    
 devtools      1.9.1   2015-09-11 CRAN (R 3.2.0)                    
 digest        0.6.8   2014-12-31 CRAN (R 3.2.0)                    
 ggplot2     * 2.0.0   2015-12-18 CRAN (R 3.2.3)                    
 gridExtra     2.0.0   2015-07-14 CRAN (R 3.2.0)                    
 gtable        0.1.2   2012-12-05 CRAN (R 3.2.0)                    
 htmltools     0.3     2015-12-29 CRAN (R 3.2.3)                    
 htmlwidgets   0.5     2015-06-21 CRAN (R 3.2.0)                    
 httr          1.0.0   2015-06-25 CRAN (R 3.2.0)                    
 jsonlite      0.9.19  2015-11-28 CRAN (R 3.2.2)                    
 magrittr      1.5     2014-11-22 CRAN (R 3.2.0)                    
 memoise       0.2.1   2014-04-22 CRAN (R 3.2.0)                    
[...]

PR #523, opened 2016-03-18
solution for #463 plotly in ioslides

timelyportfolio:
This pull represents the most minimal solution possible to #463, so tooltips will appear in `ioslides` presentations.  I tested with various examples in different contexts and did not notice any harmful regressions or side effects.


cpsievert:
Thanks, this looks good. Some tests are failing because you don't have my credentials :P

cpsievert:
In the future, would you mind sending pull requests from this remote instead of your fork?

timelyportfolio:
Oh, makes sense.  Sorry and will do going forward.

cpsievert:
I wonder if the credential issue will resolve itself in that case...would you mind resending from the ropensci remote?

PR #530, opened 2016-03-22
protect against a null key in selection

cpsievert:


inst/htmlwidgets/plotly.js
```diff
@@ -50,7 +50,7 @@ HTMLWidgets.widget({
           var tr = x.data[pt.curveNumber];
           // add on additional trace info, if it exists
           attachKey = function(keyName) {
-            if (tr.hasOwnProperty(keyName)) {
+            if (tr.hasOwnProperty(keyName) && tr[keyName] !== null) {
               if (typeof pt.pointNumber === "number") {
                 obj[keyName] = tr[keyName][pt.pointNumber];
               } else {
```
cpsievert: I was having trouble running https://gist.github.com/walkerke/da941dac84f730adcdde locally and discovered that it was because `key` is defined, but `null` (a result of recent changes to the R client). This provides a safer check before trying to subset trace info.

@timelyportfolio I'm a JS novice, so if you see anything in `inst/htmlwidgets/plotly.js` that could be improved, please let me know :)

cpsievert:
> The message below was automatically generated after build https://travis-ci.org/ropensci/plotly/builds/117837954 

 Detected 3 differences -> 

 http://cpsievert.github.io/plotly-test-table/7aca4ec/jitter-basic
http://cpsievert.github.io/plotly-test-table/7aca4ec/scatterplots-geom_jitter
http://cpsievert.github.io/plotly-test-table/7aca4ec/scatterplots-jitter

timelyportfolio:
Makes sense.  The most painful that I found are the `undefined`, so I try to protect by insuring all are at least `NULL`/`null`.  I will do a more thorough review soon, but I think you did well :)

Issue #510, opened 2016-03-15
xlab & ylab getting cut off using ggplotly()


kidman007:
Hello,

I'm new to Plotly and am very excited to use it's flash visualizations and the tooltip. I'm running into an issue with my x & y labels getting cut off when I run the viz with ggplotly -- whereas it works fine when I just run the ggplot().

Below is the code and viz. I appreciate the help in advance!

Andrew

![image](https://cloud.githubusercontent.com/assets/12884043/13791063/5996f552-eab9-11e5-81d9-b6bdb48b97f3.png)

Here's my code:
```
testPlot <-
ggplot(TopCarePaths.Req,aes(x = calc_appears_offset, y = total.completed_sum/total.completed_and_expired_sum ) )+
  geom_point( aes( shape = is_form,  color = PrePostFlag, text = paste("Title:", title_full) ), alpha = 2/5, size = 5) + #dimensions of CarePath
  # geom_point( aes( text = paste("Title:", title_full) ) ) + # changes the tooltip
  # geom_rug(sides="b", position='jitter') + #adds rug, which counts number of points. Added so only applied to X-axis & spread out with jitter
  guides( alpha = FALSE ) + #hides unneccssary legend for hard coded transparancy
  facet_grid( organization_parent_name~. ) + #Facets graph horizontally
  scale_x_continuous( limits = c( -50, 50 ), breaks = seq( -50, 50, 5 ) ) + #controls X axis tick marks
  scale_y_continuous(labels = scales::percent) + #quickly changes y axis to %
  geom_vline( xintercept = 0 ) + #adds solid line to chart
  geom_text( aes( x = 0, label = "Surgery\n", y = .2), family = 'sans', angle=0,  color = 'gray34', size = 3, alpha = 1/5 ) + #puts text next to line for surg date
  theme_bw() + #changes to a simpler, easy to read theme
  theme( strip.background = element_blank() ) + #removes ugly gray from name area of facet
  ggtitle( 'Content Completion by Across Full CarePath' ) +
  ylab('Completion %') +
  xlab( 'Days Appears Offset from Surgery' )
ggplotly()
```

bhive01:
I'm seeing this too on my plots lately, but it appears to be related to the window size/ratio. 
Here's a MRE:

```r
require(ggplot2)
require(plotly)

randomName <- function(n) {
	random.string <- rep(NA, n)
	randomizeString <- function(x) {
		a <-sample(letters, 1, replace = TRUE)
		return(a)
[...]

PR #573, opened 2016-05-05
use Plotly axis titles if not facetted

timelyportfolio:
As reported in #510, axis titles can be positioned improperly depending on the size of the plot.  In a2aedacd0900bf6f447d72e3f174deb4d8cfbc91, axis titles were changed to annotations, so they are not duplicated and work much more similarly to `ggplot2`.  Unfortunately, `plotly` annotations are positioned based on a `[0,1]` scale with respect to the plot area.  This means the axis titles scale inversely and negatively to the plot size.  When the plot is small, the `x` axis title will be placed too high.  When the plot is large/high, the `x` axis title moves too far down sometimes out of the svg bounding box.

This pull changes to use `plotly` axis titles when the plot is not facetted.  When the plot is facetted, the `plotly` axis titles are changed to `""` in favor of the annotation approach.

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -335,6 +335,9 @@ gg2list <- function(p, width = NULL, height = NULL, tooltip = "all", source = "A
       }
       # type of unit conversion
       type <- if (xy == "x") "height" else "width"
+      # get axis title
+      axisTitleText <- sc$name %||% p$labels[[xy]] %||% ""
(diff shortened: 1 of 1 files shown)


cpsievert:
Thanks, this is a great start. Would you mind starting this same pull request from the ropensci remote?

timelyportfolio:
close in favor of #574 from ropensci remote

PR #591, opened 2016-05-23
fix evaluate

sipemu:
plotly_build has as input a list in the case of evaluate = TRUE. But there is no method plotly_build.list.


cpsievert:
Yikes, thanks for pointing this out! Fixed in 289b38b1c91f28750658f2abc6410a6435db7069

PR #593, opened 2016-05-23
ggplotly: separate facet columns by <br>

alyst:
Currently ggplotly uses ` ,` to separate `facet_grid()` columns.
The PR changes that to `<br>` to make it more compatible with ggplot (every column on its own line).
It would also help with overlapping row/column long labels.

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -485,7 +485,7 @@ gg2list <- function(p, width = NULL, height = NULL, tooltip = "all", source = "A
     if (has_facet(p)) {
       col_vars <- ifelse(inherits(p$facet, "wrap"), "facets", "cols")
       col_txt <- paste(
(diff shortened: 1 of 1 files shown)


cpsievert:
Thanks @alyst, this looks promising.

Would you mind bringing this up-to-date with master? With ae6b073d0f99f32f6b58dc4247d5c38810f9bb3e, the tests should no longer report false positive test errors, and we can see a more informative test reporter. 

It would also be awesome if you could write a test with multiple line breaks to demonstrate the difference!

alyst:
@cpsievert Do you mean PR #471 ? There it's also about line breaks :) I've just rebased that one and updated the "group" tests. I can rebase this PR as well.

cpsievert:
Thanks! I'm afraid #471 might not get merged. I'm working on a big rewrite locally right now (which will fix this problem) and I'll start a pull request in the next few days.

PR #594, opened 2016-05-23
Resolve directory name change in examples

McClellandKemp:


cpsievert:
Thanks @McClellandKemp! Could you please run `devtools::document()` as well?

PR #597, opened 2016-05-23
A better way to import/export pipe

talgalili:
I found this fix here:
https://github.com/hadley/dplyr/blob/master/R/utils.r
And it works better (i.e.: will save us from pointless warnings with other packages who export it in the old way)


cpsievert:
Thanks @talgalili! Could you please run `devtools::document()` as well?

talgalili:
I updated the roxygen.

Cheers,
Tal

PR #613, opened 2016-05-30
No '.embed' extension for Jupyter embedded pages

### LINKED ISSUE #581: plotly_POST not embedding plot as expected
So this is a little weird, and I haven't been able to figure it out after reading the documentation ([1](https://plot.ly/r/), [2](https://plot.ly/r/file-options/), [3](https://plot.ly/rest/), [4](https://plot.ly/ggplot2/)).

I generated a plot using ggplot2, which I then embed in a R Markdown document published as HTML on the web (Jekyll blog, but that's beside the point).

```
```{r lnb-160411-quality-fits, echo=FALSE, results="asis"}
p <- ggplot() + ...
plotly_POST(p, filename = "lnb-160411-poor-fits", fileopt = "overwrite", sharing = "public")
```
This produces the following HTML:

```
<iframe src="https://plot.ly/~chepec/151" width="800" height="600" id="igraph" scrolling="no" seamless="seamless" frameBorder="0">
</iframe>
```
Which *does not* produce the expected result, i.e., an embedded plot, but instead **embeds the entire Plotly page** (see below).
google chrome

I confirmed that the URL in the iframe above is in fact the output of `plotly_POST`, 

```
> x <- plotly_POST(p, filename = "lnb-160411-poor-fits", fileopt = "overwrite", sharing = "public")
No encoding supplied: defaulting to UTF-8.
Success! Modified your plotly here -> https://plot.ly/~chepec/151
> str(x)
List of 5
 $ url     : chr "https://plot.ly/~chepec/151"
 $ message : chr ""
 $ warning : chr ""
 $ filename: chr "lnb-160411-poor-fits"
 $ error   : chr ""
 - attr(*, "class")= chr "figure"
```
and I think I can confirm that Plotly shows the embed *only* when appending `.embed` to the URL (the HTML below produces the expected embedded plot):

```
<iframe src="https://plot.ly/~chepec/151.embed" width="800" height="600" id="igraph" scrolling="no" seamless="seamless" frameBorder="0">
</iframe>
```

I think this must be an issue with `plotly_POST`? It should return the URL of an embed plot, not the whole page. By the way, this behaviour must have changed quite recently because the very same `plotly_POST(...)` call produced a properly embedded plot just a month ago.

Nikolai-Hlubek:
The fix for #581 by adding '.embed' to plotly_iframe breaks compatibility with Jupyter. 

The '.embed' extension is not added in all necessary places: New files are saved without the extension, and than they can't be loaded, because the iframe expects the extension. 

However fixing this would still not work with Jupyter, as Jupyter does not render files with the extension '.embed'. It only renders if the extension is html otherwise it shows the contents as text.

This commit is a possible suggestion that allows fix #581 to remain valid and saves graphs for Jupyter without '.embed' extension.

### DIFF
--- a/R/print.R
+++ b/R/print.R
@@ -103,7 +103,7 @@ knit_print.figure <- function(x, options, ...) {
   knitr::asis_output(iframe)
 }
 
-#' Embed a plotly figure as an iframe into a IPython Notebook
+#' Embed a plotly figure as an iframe into a Jupyter Notebook
 #' @param x a plotly object
 #' @param width attribute of the iframe. If \code{NULL}, the width in
 #' \code{plot_ly} is used. If that is also \code{NULL}, '100\%' is the default.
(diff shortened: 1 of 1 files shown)


cpsievert:
Thanks!

PR #619, opened 2016-06-03
Fix the axes layout for facet grid/wrap.

alyst:
The PR fixes the labels of axes/anchors if the corresponding indices are >= 10 (otherwise the data/gridlines in some panels are missing). 

For a test one can try
```R
                   a = sample(1:20, 1000, replace=TRUE) ) ) +
  geom_point(aes(x=x, y=y)) + facet_wrap(~ a, scales = "free_x" )
ggplotly(p)
```

The PR fixes the behaviour for `scales="free"` and `scales="free_x"` case. For `scales="free_y"` case it fixes the missing panels, but the X axes at the bottom are still missing.

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -212,10 +212,11 @@ gg2list <- function(p, width = NULL, height = NULL, tooltip = "all", source = "A
     }
   }
   # format the axis/anchor to a format plotly.js respects
-  panel$layout$xaxis <- paste0("xaxis", sub("1", "", panel$layout$xaxis))
-  panel$layout$yaxis <- paste0("yaxis", sub("1", "", panel$layout$yaxis))
-  panel$layout$xanchor <- paste0("y", sub("1", "", panel$layout$xanchor))
-  panel$layout$yanchor <- paste0("x", sub("1", "", panel$layout$yanchor))
+  plotlyjs_axis_label <- function(prefix, ixs) paste0(prefix, ifelse(as.integer(ixs) > 1, ixs, ""))
+  panel$layout$xaxis <- plotlyjs_axis_label("xaxis", panel$layout$xaxis)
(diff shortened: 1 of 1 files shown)


alyst:
Also fixed the placement of X axes when then last row is not complete, the axes is attached to the lowest subplot in the column, not the last row of the whole plot. Otherwise plotly.js draws fake empty subplot to attach the axis to.

cpsievert:
Thanks @alyst. I'm closing this, but you're contributions are much appreciated, and have been quite useful for identifying issues. Some of these ideas have been incorporated into #654.

alyst:
@cpsievert i've looked at the code of your PR but have no time to test it. I think there could still be issues with axes anchoring if you have e.g `xaxis21`. Your code would transform it to `xaxis2`. Also i haven't understood whether you have a fix for axes anchoring if the last row is incomplete.

alyst:
The issue with >10 indices of axes is fixed, but the issue with X axis placement on the incomplete last row is still in the plotly's master.

*ggplotly* handles the last row differently from *ggplot2*: `ggplot()` attaches the X axis to the last **plot** in the column, whereas `ggplotly(ggplot())` always puts it to the last **row**, even when there are no plots at that position. IMHO the *ggplot2* behaviour is better -- there are no "floating" axes, the axes are always attached to the plots.

I've rebased this PR to the master to fix the issue. Could you please reopen the PR?

There's one testcase affected by the fix: "facet_wrap-free_y-2" (it also demonstrates the difference between ggplot and current ggplotly behaviours). I guess I also need to fix the reference .png, but I'm not sure how to properly do it.

cpsievert:
Hi @alyst, for some reason I can't reopen...would you mind starting a new PR?

Also, thanks for the recent burst in PRs!

Issue #522, opened 2016-03-17
Missing Facet Panels


bhive01:
I'm playing around with some different viz at the moment and when I try to facet_wrap I get empty panels in places where I don't expect them. Also there are loads of extra occurrences of the color factors in the legend.

![screen shot 2016-03-17 at 4 29 29 pm](https://cloud.githubusercontent.com/assets/394973/13864308/9830a114-ec5d-11e5-8983-68e52b3466af.png)

```r
library(ggplot2)
library(plotly)
[...]

PR #620, opened 2016-06-03
Fix trace processing if number of subplots >= 10

alyst:
Another fix for multi-subplots case.

Test case (single aesthetic bound to trace):
```R
                   a = sample(1:20, 1000, replace=TRUE),
                   b = factor(sample(1:5, 1000, replace=TRUE))) ) +
  geom_point(aes(x=x, y=y, color=b)) + facet_wrap(~ a )

ggplotly(p)
```

With multiple aesthetics per trace:
```R
                   a = sample(1:20, 1000, replace=TRUE),
                   b = factor(sample(1:3, 1000, replace=TRUE)),
                   c = factor(sample(1:3, 1000, replace=TRUE))) ) +
  geom_point(aes(x=x, y=y, color=b, shape=c)) + facet_wrap(~ a )

ggplotly(p)
```

### DIFF
--- a/R/layers2traces.R
+++ b/R/layers2traces.R
@@ -36,9 +36,10 @@ layers2traces <- function(data, prestats_data, layout, p) {
     d <- datz[[i]]
     # always split on discrete scales, and other geom specific aesthetics that
     # can't translate to a single trace
-    split_by <- c(split_on(d), names(discreteScales))
+    split_by <- unique(c(split_on(d), names(discreteScales)))
+    if (length(split_by)>0) split_by <- paste0(split_by, "_plotlyDomain")
     # always split on PANEL and domain values (for trace ordering)
-    split_by <- c("PANEL", paste0(split_by, "_plotlyDomain"))
+    split_by <- c("PANEL", split_by)
     # split "this layers" data into a list of data frames
     idx <- names(d) %in% split_by
     # ensure the factor level orders (which determies traces order)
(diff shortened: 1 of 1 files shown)


alyst:
Previously the `split_by` name components were excluded from the trace name generation only if they were followed by the "@%&". That concealed the bug in estimating how many scales should be excluded (`split_by` scales have to be used, not the original data columns).

R/layers2traces.R
```diff
@@ -55,16 +56,20 @@ layers2traces <- function(data, prestats_data, layout, p) {
     trs <- Map(geom2trace, dl, paramz[i], list(p))
     # are we splitting by a discrete scale on this layer?
     # if so, set name/legendgroup/showlegend
-    isDiscrete <- names(d) %in% paste0(names(discreteScales), "_plotlyDomain")
+    isDiscrete <- names(d) %in% (if (length(discreteScales) > 0) { paste0(names(discreteScales), "_plotlyDomain") } else as.character())
     if (length(trs) > 1 && sum(isDiscrete) >= 1) {
```
cpsievert: I'd prefer 

```r
isDiscrete <- names(d) %in% paste0(names(discreteScales), "_plotlyDomain") %||% ""
```

R/layers2traces.R
```diff
@@ -36,9 +36,10 @@ layers2traces <- function(data, prestats_data, layout, p) {
     d <- datz[[i]]
     # always split on discrete scales, and other geom specific aesthetics that
     # can't translate to a single trace
-    split_by <- c(split_on(d), names(discreteScales))
+    split_by <- unique(c(split_on(d), names(discreteScales)))
+    if (length(split_by)>0) split_by <- paste0(split_by, "_plotlyDomain")
     # always split on PANEL and domain values (for trace ordering)
```
cpsievert: @alyst I prefer using white space in logical expressions

R/layers2traces.R
```diff
@@ -55,16 +56,20 @@ layers2traces <- function(data, prestats_data, layout, p) {
...
-        if (length(x) > 1) paste0("(", paste0(x, collapse = ","), ")") else x
+        n_x <- length(x)
+        if (n_x > n_ignore + 1) {
+          paste0("(", paste0(x[(n_ignore+1):n_x], collapse = ","), ")")
+        } else if (n_x > n_ignore) {
+          x[[n_ignore+1]]
+        } else {
+          NA_character_
+        }
       }, character(1))
```
cpsievert: @alyst I don't completely follow this logic. Could you explain a little? Also, this PR will definitely need some tests before it gets merged. Here would be a good example to test https://github.com/ropensci/plotly/issues/522

alyst: Instead of removing the first ignored scales using `sub()` in the loop, I've changed the code to directly split the trace ids by "@%&" and then ignore the first `n_ignore` chunks (non-discrete scales). Then the visible trace names are composed as before: if there's more than one scale left after `n_ignore`, the trace name is like `(a, b)`, otherwise the remaining scale name is used "as is".
In the original code there was an error in the `sub()` regex that manifested itself when the length of the non-discrete scale id was more than 1 symbol. So I thought it would be cleaner and safer to avoid preprocessing trace Ids.

alyst: Adding the test for #522 definitely makes sense. Do you think of any other tests that should be added before merging this PR?

cpsievert:
Thanks @alyst. I'm closing this, but you're contributions are much appreciated, and have been quite useful for identifying issues. Some of these ideas have been incorporated into #654.

PR #622, opened 2016-06-05
Fix the margins of subplots

alyst:
In master the subplot margins are *subtracted* from the given subplot width/heights. Since top-left subplot should start in (0,1) and bottom-right should end at (1,0), the left margin is not subtracted from the leftmost subplots and the right margin -- from the rightmost. The same for heights. As the result the proportions of the plots drawn differ from the ones given to `subplot()`: in comparison to outer subplots the inner ones are more narrow than they should be. See e.g.:
```R
                   a = sample(1:5, 1000, replace=TRUE),
                   b = factor(sample(1:3, 1000, replace=TRUE)),
                   c = factor(sample(1:3, 1000, replace=TRUE))) ) +
  geom_point(aes(x=x, y=y, color=b, shape=c)) + facet_wrap(~ a )
ggplotly(p)
```

The PR changes how the `get_domains()` calculates the subplot positions. Instead of subtracting the margins from the subplot widths/heights, it rescales the width/heights so that the sum of margins and width/heights does not exceed 1.0 (the margins are not scaled since they are calculated wrt ticks/labels sizes). This approach preserves the relative subplot proportions. It looks like it also fixes the `free_y` case in #619.

There are 2 open issues related to this PR:
1. For `get_domains()` there's no difference between left/right, only the left+right sum does matter. The same for top/bottom. So instead of expecting 4-element `margins` vector, `get_domains()` could be simplified to expect 2-element `spaces` vector.
2. Margins take into account the size of labels converted from pixels to (0,1) range. Since the plot size on the server side (used for this conversion) could be different from the actual plot size on the client, the resulting layout might be incorrect. The current workaround is to double the margin sizes (see the comment in `ggplotly()`) to avoid subplot overlaps. But certain combinations of client/server plot sizes result in overly large margins and reduced subplots. Would be nice if at least upon redraws the server can get the actual plot dimensions.


alyst:
I've just tried and with the current `plotly` version the subplot margins issue is still there (margins too wide, subplots have uneven sizes), so I've rebased the PR.

nlooije:
I patched the latest release with these changes and the problem with the subplot margins seems to have been fixed.
Would be nice to have them merged into master some time soon

Issue #600, opened 2016-05-24
Make ggplotly() work better anytime layout.axisid.ticktext is supplied


talgalili:
This affects issue #599 .

Minimal example:

``` r
library(ggplot2)
data(diamonds)
diamonds$cut <- paste("Super Dee-Duper extra long",as.character(diamonds$cut))
p <- qplot(cut,carat,data=diamonds,geom="boxplot") + 
  theme(axis.text.x = element_text(angle = 90, hjust = 1))
p
```

![image](https://cloud.githubusercontent.com/assets/976006/15492522/c7b935c0-2184-11e6-86a1-acecd0cd2d7c.png)

``` r
library(plotly)
ggplotly(p) # right most label is cut
```

![image](https://cloud.githubusercontent.com/assets/976006/15492532/dd42c712-2184-11e6-9e78-6a6e6a16436f.png)

cpsievert:
I believe the new-ish auto-margin support fixed this issue.

PR #623, opened 2016-06-06
RFC: virtual width/height for ggplotly()

alyst:
This is the proof-of-concept workaround for the layout problems of ggplotly() (see e.g. #622).
The problem is that when converting from absolute sizes of e.g. labels ("inches", "mm" etc) to relative sizes ("npc") ggplotly uses `grid::convertXXX()` functions that use the dimensions of the active device, which may not match the size of the plotly plot on the client.
The PR provides at least the way to control the size of the active device on the server side. It introduces `virt.width`/`virt.height` arguments to `ggplotly()` that serve as the expected dimensions of the client plot. If set to numeric values, `ggplotly()` creates the viewport of the specified dimensions (in "points"), so that all conversions to/from "npc" are done relative to this viewport. `virt.width/height` do not affect the size of the resulting plot, neither they interfere with the `width/height` (which might be set to something not numerical as e.g. "100%").

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -31,14 +31,19 @@
 #'  ggplotly(viz)
 #' }
 #'
-ggplotly <- function(p = ggplot2::last_plot(), width = NULL, height = NULL,
+ggplotly <- function(p = ggplot2::last_plot(),
+                     width = NULL, height = NULL,
+                     virt.width = ifelse(is.numeric(width), width, 1024),
+                     virt.height = ifelse(is.numeric(height), height, 768),
                      tooltip = "all", source = "A", ...) {
   UseMethod("ggplotly", p)
 }
 
 #' @export
-ggplotly.ggmatrix <- function(p = ggplot2::last_plot(), width = NULL,
-                              height = NULL, tooltip = "all", source = "A", ...) {
+ggplotly.ggmatrix <- function(p = ggplot2::last_plot(),
+                              width = NULL, height = NULL,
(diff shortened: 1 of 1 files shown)


talgalili:
Hi @alyst 
Does this help solve #600 ?

Also, would this make the margin in the following work (without the need to use layout):
```r
library(heatmaply)
heatmaply(mtcars)
```

?

Thanks,
Tal

alyst:
@talgalili It might help, although (running `plotly` with the other layout patches I've recently submitted) I cannot reproduce ticks labels clipping.

I don't know if the PR would help `heatmaply`. Currently virtual viewport workaround is implemented for `ggplotly()` variant that accepts `ggplot` objects. It's very easy to extend the PR to the other plot types, but if `heatmaply` is not inherited from `ggplot` the current version would do nothing.

talgalili:
@alyst [heatmaply](https://github.com/talgalili/heatmaply/blob/master/R/heatmaply.R) works using ggplotly (using "geom_tile"), so I think it might help. If you'll get to check it (and it helps) - please let me know.

Best,
Tal

cpsievert:
This is a nice idea, thank @alyst, sorry for the delay. 

Why do u set defaults of 768 and 1024 for height/width?

cpsievert:
I don't think we need to add  additional arguments... see 9afbb5da88dd510c3aa44c855d8fb72c4e4b718e

Thanks for the idea @alyst!

PR #627, opened 2016-06-16
fix update script and updated plotlyjs to 13.0

yankev:
I changed the path for file copy because I ran `plotlyjs.R` from within its directory.
For Issue #626.

PR #628, opened 2016-06-16
Major update to `plot_ly()`

cpsievert:
@jackparmer @timelyportfolio 

This is a _major_ update to the `plot_ly()` interface that addresses a number of concerns raised by @hadley / @jcheng5 / others. See the [NEWS](https://github.com/ropensci/plotly/blob/fix/nse/NEWS) file for a summary. My goal is to have this on CRAN before I fly out to SF on the 23rd. Here is a list of stuff that should be addressed before submitting:

- [x] Get a working version of `add_polygon()`/`add_ribbon()`. BTW, if anyone has ideas for other `add_*()` functions (to do things that aren't easy/available or slow in ggplot2) I'm all ears!
- [x] Add tests/documentation
- [x] Fix grouping when there is multiple traces, for instance:

```r
g <- expand.grid(visit = 1:10, id = 1:50, cohort = c("A", "B"))
g$response <- rnorm(nrow(g))
g %>%
  group_by(id) %>%
  plot_ly(x = ~visit, y = ~response, color = ~cohort, opacity = 0.5)
```

I also think that `color` should be deprecated and split into `stroke`/`fill`, but that should probably wait for now.


cpsievert:
@timelyportfolio it looks like resetting zoom within an R markdown doc is broken on this branch. Could you check out #627 for me and see if it's just v1.13.0 that breaks it?

timelyportfolio:
This will be really nice.  I am working through examples and tests now.  The first issue I noticed is in the `?plot_ly`.

timelyportfolio:
this works 

```
plot_ly() %>% add_data(economics) %>% add_trace(x = ~date, y = ~pce) %>% add_trace(x = ~date, y = ~log(pce,base=10))
```

so I'm struggling to understand why the other does not.

R/plotly_build.R
```diff
@@ -0,0 +1,373 @@
...
+    x[lengths(x) > 0]
+    
+  }, p$x$layoutAttrs, names(p$x$layoutAttrs))
+  
+  # get rid of the data -> layout mapping and merge all the layouts
+  # into a single layout (more recent layouts will override older ones)
+  p$x$layoutAttrs <- NULL
+  p$x$layout <- modify_list(p$x$layout, Reduce(modify_list, layouts))
+  
+  # If type was not specified in plot_ly(), it doesn't create a trace unless
```
timelyportfolio: it seems this explains https://github.com/ropensci/plotly/pull/628#issuecomment-226785704, but I don't understand the logic.   Could you elaborate on this design?

timelyportfolio: If I comment out, then both the smoother and raw line are plotted, so this does seem to explain what is happening, but I don't understand why.

cpsievert:
@timelyportfolio similar to `ggvis()`/`ggplot()`, I'd like to move to a system where `plot_ly()` _initializes_ a visualization, and defines global attributes/properties, but doesn't necessarily define a trace (i.e., layer) itself. That way, this would create a single trace, rather than 2:

```r
plot_ly(economics, x = ~date, y = ~pop) %>% add_lines()
```

If no `add_trace()`-like functions are used, we essentially add it at print time (so that these would also create the same result).

```r
plot_ly(economics, x = ~date, y = ~pop, type = "scatter")
plot_ly(economics, x = ~date, y = ~pop)
No trace type specified. Using the 'scatter' default.
```

Unfortunately, this does introduce some confusing (backwards-incompatible) changes. Take the first example on `plot_ly()`. In the past, this would create two traces:

```r
plot_ly(economics, x = ~date, y = ~uempmed) %>% 
  add_trace(y = ~fitted(loess(uempmed ~ as.numeric(date))))
```

But, given this new philosophy (which I think is better and more consistent with grammar of graphics systems), that would create one trace. This is now the preferred way to express that plot:

```r
plot_ly(economics, x = ~date) %>% 
  add_lines(y = ~uempmed) %>% 
  add_lines(y = ~fitted(loess(uempmed ~ as.numeric(date))))
```

R/layout.R
```diff
@@ -0,0 +1,49 @@
...
+
+#' @export
+layout.plotly <- function(p, ..., data = NULL) {
+  p <- add_data(p, data)
+  attrs <- list(...)
+  if (is.null(p$x$cur_data)) {
+    nlayouts <- length(p$x$layout)
+    p$x$layout <- c(p$x$layout, attrs)
+  } else {
+    p$x$layoutAttrs[[p$x$cur_data]] <- attrs
```
timelyportfolio: This will overwrite any previous defined `layout`.  Is this intentional?

```
data(economics, package = "ggplot2")
# basic time-series plot
p <- plot_ly(economics, x = ~date, y = ~uempmed)
# add a loess smoother
p2 <- add_trace(p, y = ~fitted(loess(uempmed ~ as.numeric(date))))
# add a title
p3 <- layout(p2, title = "Median duration of unemployment (in weeks)")
# change the font
layout(p3, font = list(family = "Courier New, monospace"))
```

The `title` is lost.

image

cpsievert: Whoops, no, not intentional, I'll fix it, thanks!

cpsievert: yea, working on that as we speak :)

timelyportfolio:
@cpsievert I have noticed a lot of examples follow the pattern

```
NULL %>% plotly()
```

Am I correct that this is no longer necessary?

timelyportfolio:
@cpsievert,  why does this work?

```
iris %>%
 group_by(Species) %>%
 plot_ly(x=~Petal.Width,y=~Sepal.Length,color=~Species,colors="Set1")
```

image

but not

```
iris %>%
  group_by(Species) %>%
  plot_ly(
    x=~Petal.Width,y=~Sepal.Length,
    color=~Species,colors="Set1"
  ) %>%
  add_markers()
```

image

cpsievert:
Your last example should work, but currently doesn't (that's what I meant by the 3rd checklist point above).

Other than `?add_data`, I'm not aware of examples that use a `NULL %>% plot_ly()` pattern. The first argument to `plot_ly()` defaults to `NULL`, so it's redundant, and should probably be removed

PR #671, opened 2016-08-07
fix object 'rating' not found

harryprince:
I just change the rating parameter prefix. it works.


cpsievert:
I don't see why it doesn't already work

PR #728, opened 2016-09-26
Add `source` argument to both the source plot and `event_data` function

nielsenmarkus11:
The purpose is to avoid error messages and confusion if and when someone clicks a point on the scatter plot [online.](https://plot.ly/r/shinyapp-linked-click/)


cpsievert:
Thanks!

PR #830, opened 2016-12-15
preventing too many active WebGL contexts

AleksandrIanevski:
Plotly makes one WebGL context per graph meaning that pages with multiple plotly WebGL graphs are exposed to the browser's context cap.  https://github.com/plotly/plotly.js/issues/337
So it is better to redraw the graph, rather than create a new one.
Thanks


cpsievert:
Nice, thanks!

PR #882, opened 2017-02-16
add codecov badge

steven2249:


cpsievert:
Thanks...I believe we need a codecov account moving to #963

PR #938, opened 2017-04-10
An attempt at a more general animation API

cpsievert:


cpsievert:
Also, the point of the API is to be able to transition between _different_ trace types...the current API (i.e., `frame`) assumes the animation has a fixed trace type

rreusser:
What does it do visually as it transitions types? That is, what does it even mean to transition a bar graph into a scatter trace? A long-running goal is to smooth out those things so they at least do *something* nice/reasonable.

cpsievert:
> What does it do visually as it transitions types?

I'm hoping plotly.js can work out those details ;)

More seriously, this is just an attempt to make it easier to "animate" from one plot to another (i.e., it's ok if a redraw is required)

PR #979, opened 2017-04-26
What are the possible args for "off"?

talgalili:
The error message says it should be "plotly_relayout"

```
> highlight(ggplotly(p), on = "plotly_selected", off = 'plotly_unhover', persis = FALSE)
Error in match.arg(off, off_options) : 
  'arg' should be one of “plotly_doubleclick”, “plotly_deselect”, “plotly_relayout”
> 
```


cpsievert:
Thanks!

PR #984, opened 2017-04-28
Developer Experience Audit

RichardLitt:
This is a documentation checkup. These changes were made:

- Update Plotly's license name to match what they use on their own repos;
- Update the CoC to the most recent Contributor Covenant version, with an email account for people to check;
- Standardize the README according to [RichardLitt/standard-readme](https://github.com/RichardLItt/standard-readme). Some of these changes are sugar, but a few have reasons - badges below heading for readability on  non-Markdown rendered interfaces, for instance.
- Added a Maintainers section to encourage others to want to be maintainers, and to provide an explicit point of reference. I added @cpsievert there - let me know if we should add more!
- Added a license section to the README.

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,25 +1,74 @@
-# Contributor Code of Conduct
+# Contributor Covenant Code of Conduct
 
-As contributors and maintainers of this project, we pledge to respect all people who 
-contribute through reporting issues, posting feature requests, updating documentation,
-submitting pull requests or patches, and other activities.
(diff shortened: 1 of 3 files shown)


cpsievert:
Thanks, I'm pretty sure our conduct conforms to @ropensci's standards...perhaps @karthik can give us his two cents on this.

RichardLitt:
Cool. The updates are worthwhile, I think. I pulled it from the same source as the original one.

PR #1123, opened 2017-09-20
Update plotly.R

Tim113:
Only give the plotly widget and id, that will throw a warning in shiny, if it is needed.


cpsievert:
In other words, install the dev version of plotly and the warning should go away

Issue #1133, opened 2017-10-05
Bubble plot miscalculates size when all points within a color group are the same size


brianstamper:
This bug can occur when creating a bubble plot where a discrete variable determines color and a continuous variable determines size. If all of the points within a given color are the same size (or if there is only one point within that group) then the bubbles are sized incorrectly. When using `sizemode = 'diameter'` this issue does not occur, it only happens with the default `sizemode = 'area'`. Based on the experiments shown below it appears that these problematic points are having the size computed as diameter instead of area in either case, as they appear to have the same size in both modes.

I first asked about this on SO at (https://stackoverflow.com/questions/46571892/r-plotly-scatter-plot-marker-size-appears-to-depend-on-color-when-color-is-discr)

I did not include examples where color is based on a numeric or integer instead of factor or character, but I can attest this bug _does not_ occur in those cases. (In cases where the legend is a gradient instead of discrete values.)

Possibly related to https://github.com/ropensci/plotly/issues/755 about size of bubbles in the legend?

```r
library(plotly)

# In these plots I will set x, y, and size to the same variable, so the
# largest bubbles should appear on the top-right and smallest on the 
# bottom-left. The bug still occurs if you set x and y to other variables,
# but it becomes less obvious to see the phenomenon being described.

# Make a version of mtcars where 'carb' is a discrete variable. Here I
# make it a factor, but as.character() also works to demonstrate this bug.
my_mtcars <- mtcars
my_mtcars$carb <- as.factor(my_mtcars$carb)

# With the default sizemode = 'area', we have the bug, for the 
# points corresponding to my_mtcars$carb %in% c(6, 8)
plot_ly(my_mtcars,
        x = ~wt,
        y = ~wt,
        size = ~wt,
        color = ~carb,
        type = 'scatter',
        mode = 'markers',
        marker = list(sizemode = 'area'))  # This is actually the default
```
![plot1](https://user-images.githubusercontent.com/636124/31231424-ced71976-a9b4-11e7-96cb-49eaf94109f1.png)

```r
# Set sizemode = 'diameter' and the bug goes away
# (though I notice the problematic bubbles are the same size as
# they were in the plot with the sizemode = 'area'... hmm...)
plot_ly(my_mtcars,
        x = ~wt,
        y = ~wt,
        size = ~wt,
        color = ~carb,
        type = 'scatter',
        mode = 'markers',
        marker = list(sizemode = 'diameter'))
```
[...]

PR #1176, opened 2018-01-09
ensure size arg always maps to array, closes #1175

### LINKED ISSUE #1175: map_size doesn't map vector of length 1 to array
Turns out plotly.js marker.size yields different results for a constant vs an array of the same numbers, see for example

https://codepen.io/cpsievert/pen/RxQjyK
https://codepen.io/cpsievert/pen/qpxVKd

Therefore, when using `size` it should probably *always* map to an array

cpsievert:


cpsievert:
Actually I think the right way to proceed is slightly different -- see https://github.com/ropensci/plotly/issues/1133#issuecomment-397116217

PR #1178, opened 2018-01-11
Bump plotly.js version

cpsievert:
Things that need to be done before we can safely update:

  - [x] Integrate `selected` attribute/api with crosstalk framework.
  - [ ] Integrate `unselected` attribute/api with crosstalk framework (should `opacityDim` map to `unselected.marker.opacity`, etc?). 
(__UPDATE__: I can't think of a good use-case for changing stuff like `marker.color`, `marker.size`, or `textfont.color` for unselected marks, and in fact, I think it just leads to more confusion than usefulness, so I'm not likely going to add integration) 
  - [x] Enable persistent selection via 'shift' key (this should work for all valid event types, not just `plotly_selected`)
  - [ ] Does it make sense for `selectedpoints` to set an initial crosstalk selection (like `defaultValues` arg in `highlight()`)?
(__UPDATE__: It's definitely not straight-forward how one would use `selectedpoints` in conjunction with crosstalk, and adding support wouldn't enable anything that can't already be done via `defaultValues` in `highlight()`, so I'm not sure the dev time this would require is necessarily worth it) 
  - [x] Wait until fix for https://github.com/plotly/plotly.js/issues/2282 has been included in an official release


daattali:
Two more small issues with this PR (they might automatically get fixed when you fix the bigger selection issue)

1. When the initial dragmode is zoom and I manually click on lasso selection, the plot stretches horizontally a bit (see GIF below)

2. When the plot initially loads it looks and scales correctly. When the window is resized to be narrower/wider, the points are not longer kept in the proper location. In the GIF below, after making the windows narrower, many points that were originally positive in the X axis moved over to be to the left of the X axis

cpsievert:
@daattali I've reported a minimal example here -- https://github.com/plotly/plotly.js/issues/2282

cpsievert:
@daattali re-installing this PR should fix at least your 2nd point, let me know if you find anything else!

daattali:
Thanks. I'll take your word for it, I'll test it out when PR is ready/merged

daattali:
The second issue I raised is indeed fixed. The first isn't. But a much bigger issue is that selecting points as you did in my app seems to be very broken:

plotlybugs

cpsievert:
@daattali thanks, that's odd that the positions change (I wasn't able to replicate that bit), but there does indeed seem to be a problem with Plotly.addTraces() that could explain the lack of color change -- https://github.com/plotly/plotly.js/issues/2298

daattali:
Try the app that you helped me work on. When I install plotly from 5e4da510b020093fc573d3ba348c27caa6a9185b it works perfectly fine, when I install plotly from today, selection is messed up

Issue #1104, opened 2017-08-19
Plotting large data sets


Nikolai-Hlubek:
Basically this is another go at bug number #822. 

R Plotly is not very good in handling large numbers of data. I'm using R in Jupyter and when I plot a line >1e5 points, which I do not consider to be much, I get a pandoc conversion error. 

Python Plotly can easily handle a line with 1e6 points.

Ok, so this might be an issue with pandoc, so I played around a bit and removed the pandoc dependency by doing the html handling manually. 
See the attached zip. This works and I get plots with 2e5 points in a reasonable amount of time. 
I'm actually convinced the way I'm doing it with the precompiled header which is just inserted into the html, instead of using Pandoc to create the header is the way to go.

However at 3e5 points the routine slows down significantly. The reason is the long time spent in plotly_built. So I'm stopping here and asking for help: Help! 
Is there anything that can be done to speed this routine up? A million points is really not a lot nowadays and should be possible with R plotly.

[embed_R_without_PANDOC.zip](https://github.com/ropensci/plotly/files/1236694/embed_R_without_PANDOC.zip)

PS: 
With a bigger box I is possible to plot larger datasets. Therefore for a fair comparison I'm using vagrant to create a virtual machine with python, R, plotly and jupyter:

https://github.com/Nikolai-Hlubek/vagrant-centos7-jupyter

msummersgill:
What version of plotly are you using? When I try to reproduce, I'm seeing several other bottlenecks as the primary constraint, not the `plotly_build()` function.

I took a subset of your function to render using pre-compiled headers and reduced it down to only saving the html as below.

```r
plotly_save_html <- function(x, width = NULL, height = NULL, file = NULL) {
  
  precompiled_html_header <- readRDS("precompiled_html_header.rds")
  p <- plotly_build(x)
  html <- htmlwidgets:::toHTML(p)
  rendered <- htmltools::renderTags(html)
  
  id <- substr(rendered$html, 10, 20)
  linebreak <- regexpr('\n', rendered$html)
  rendered$htmlwidget <- paste0("  ", substr(rendered$html, 1, linebreak))
  
  rendered$htmlwidget_sizing <- paste0(
    "<script type=\"application/htmlwidget-sizing\" data-for=\"",
    id,
[...]

PR #1232, opened 2018-04-09
Efficiently plot large datasets / get rid of pandoc dependency - Addresses #1104

Nikolai-Hlubek:
Addresses issue: _Plotting large data sets #1104_

I have been running a custom version of plotly for several months at my company now and we did not experience any issues.

Changes:

 - I use precompiled headers containing the plotly javascript code.
 - I do not use htmlwidgets and thus pandoc but xml2. This gets rid of pandoc as external dependency and drastically increases the number of points a figure can contain.

 - I allow embed_notebook with external save files again.
This is important because otherwise the notebooks get to big. If you are working across a slow internet connection, a notebook with 100MB cannot be efficiently saved. If you change one line of code and save, all figures need to be transferred again.


cpsievert:
@Nikolai-Hlubek thanks for the PR! At the moment, this solution doesn't seem general or maintainable enough for me to accept (e.g., how do I update the precompiled header when I update/add/edit dependencies? Why `writePlotToFileXML2()` is needed instead of using `htmlwidgets::saveWidget(p, file, selfcontained=FALSE)` which is essentially `htmltools::save_html()`?)

I'd prefer to use `htmltools::save_html()`, but the problem is how to import/manage external dependencies. IIRC, for security reasons, it's not possible to spin up a local file server in Jupyter (that the HTML dependencies can reference), correct? If I'm wrong, and it is possible, that would be my preferred solution. Another possible approach is to base64 encode dependencies "by hand" using `base64enc` (or similar) package.  The hard thing with this approach is coming up with a system that recognizes and prevents redundant dependencies. I've actually tried to come up with a general solution for this in repr/IRdisplay which would solve the problem in general for htmlwidgets. For some reason, I never submitted the PR, so I can't remember if it ever worked, but here it is: https://github.com/cpsievert/repr/commit/38bf1e83319f091ec3b4f97570a21199009a50e1 

PS. Are you using Jupyter Notebook? If yes, have you tried printing plotly graphs without `embed_notebook()` and noticed any differences?

PSS. Have you tried using pandoc 2.0 or higher?

Nikolai-Hlubek:
@cpsievert I tried to explain the problem in issue #1104. In short _htmlwidgets::saveWidget(p, file, selfcontained=FALSE)_ at some point calls _pandoc_. This crashes already with a small amount of points. The amount of points it can handle is dependent on your systems memory. This is why I provided a vagrant virtual machine for a fair comparison. 
In my test 
 - python plotly could handle 1e6 points. 
 - R plotly crashed at 2e5 points.
This is problematic since I have _large_ datasets and thus cannot effectively work with plotly. Hence the need for a solution. 

So my solution is to avoid pandoc altogether. 
There is also another issue regarding speed in the current plotly R implementation. You create the plotly javascript header everytime a plot is done. In my opinion this is wasting ressources. 

Therefore my solution is as follows:
1. Generate the header once. You still need a script if you want to automatically update the header once you change dependencies. But you only need to do this once not at every plot command.
2. Rewrite the required tags using xml2. 
3. Result: No more pandoc crashes. R plotly can now handle more than 2e5 points.


As for jupyter, yes I am using jupyter notebooks. 
What you need to understand there is, that jupyter captures output and embeds it in its notebook. One empty plotly plot is currently 5MB. If I have a few plots with some data the notebook size quickly is 100MB. Clicking on save in the notebook will then write 100MB to disk. 
This is just not feasible if you have a 
 - write a line of code
 - run it
 - save it 
 - iterate
kind of workflow. Because each write operation slows you down by several seconds or so. 
On the other hand if the pictures are included as links the notebook size is a few kB and saving is super fast. 

I do not know if this is the best solution or if it has some other issues. All I can say is that it has been working for us at my company for three months now without any issues. Looking at #1104 some other people seem to have noticed this issue as well. So I just wanted to share my findings so other people can benefit.

cpsievert:
I understand the problem. `htmltools::save_html()` doesn't use pandoc, so it doesn't suffer from the large # of points issue. Also, as I mentioned, cpsievert/repr@38bf1e8 has a dependency tracker (i.e., it doesn't include more dependencies than necessary). I don't see how your solution addresses that problem -- it seems to [inject the same header on every print](https://github.com/ropensci/plotly/pull/1232/files#diff-5c326f0be068072b0a99dbbde818165aR104). It also has no way of considering other html dependencies outside of plotly

Nikolai-Hlubek:
I do not understand, why there should be the need for any other html dependencies. But the reason for this might be, that I only use a very specific subset of plotly in a very specific scenario for which my solution is enough. 

Anyway if you have a better solution already and understand the issue that is enough for me. I can hope it will be fixed eventually and in the meantime I'm fine with what I have. 
Thank you for your patience and explaining everything!

cpsievert:
> why there should be the need for any other html dependencies.

What if you want to plot multiple widgets at once? https://plotly-book.cpsievert.me/arranging-htmlwidgets.html

I've started a PR that should address most of these issue for all htmlwidgets/htmltools tags. If you'd like to help out, please test and verify it's working for your use cases -- https://github.com/IRkernel/repr/pull/89

PR #1256, opened 2018-05-16
drop ggplot2 2.2.1 support (i.e., require most recent CRAN release)

cpsievert:


DESCRIPTION
```diff
@@ -22,7 +22,7 @@ URL: https://plot.ly/r, https://cpsievert.github.io/plotly_book/, https://github
 BugReports: https://github.com/ropensci/plotly/issues
 Depends:
     R (>= 3.2.0),
-    ggplot2 (>= 2.2.1)
+    ggplot2 (> 2.2.1)
 Imports:
```
cpsievert: TODO: Should probably do ggplot2 (>= 2.3.0) 

(or whatever version they end up calling it)

cpsievert: Maybe it would be better to not have a strict version requirement in the DESCRIPTION and check the version in `ggplotly()` with an error/warning for outdated versions. That way folks can use non-`ggplotly()` stuff without having to adhere to strong ggplot2 version requirements

Issue #1270, opened 2018-05-30
Provide a way to set locale


cpsievert:
https://github.com/plotly/plotly.js/tree/master/dist#to-include-localization

PR #1271, opened 2018-05-30
add locale arg to config() which automatically adds relevant script, …

cpsievert:
…addresses #1270


cpsievert:
NOTE TO SELF:

Some locales depend on external assets -- https://codepen.io/etpinard/pen/pKvLVX

Not sure how to generally find all required assets for a given locale -- https://plotly.slack.com/archives/C07MRPNHW/p1527707166000042

PR #1326, opened 2018-08-22
quick implementation of most of ggraph/ggforce

cpsievert:


gvdr:
Was this PR ever considered? What's the situation now?

cpsievert:
@danton267 would you mind filing an issue requesting this feature? Having some `{ggraph}`/`{ggforce}` support would have a much higher ROI compared to some of the extensions you've requested here https://github.com/plotly/plotly.R/issues/created_by/danton267

moutikabdessabour:
@cpsievert I will happily work on some of the geoms

cpsievert:
@moutikabdessabour it'd be great if you could also look into converting `{ggraph}` based treemaps, so that we can avoid the potential licensing issues with converting `{treemapify}` https://github.com/plotly/plotly.R/pull/2051

moutikabdessabour:
@cpsievert currently implementing ggraph's `GeomAxisHive`

moutikabdessabour:
I have successfully deconstructed the geoms however I'm facing some `transparency` issues.

the ggplot one is a bit opaque because alpha gets rescaled to fill the 0 to 1 interval.
ggplot
 
`ggplotly` was complaining that unique alpha was bigger then one so I added `alpha` as an aes to the `split_on` function.
ggplotly

moutikabdessabour:
I implemented the ggraph's `treemap`
Screen Shot 2021-11-03 at 4 25 20 AM

Do you think I should go forward and implement all the examples found here?? https://www.data-imaginist.com/2017/ggraph-introduction-layouts/

jamesdalg:
Just want to put in my two cents that this would be an incredible addition if it somehow passes checks one day.

PR #1343, opened 2018-09-20
add support for partial updates in style(), fixes #1342

### LINKED ISSUE #1342: The style() function doesn't support partial updates
The [`Plotly.restyle()`](https://plot.ly/javascript/plotlyjs-function-reference/#plotlyrestyle) JavaScript method has long supported "partial updates" (i.e. modification of a particular property of an object, rather than the entire object). Here is a mini example:

https://codepen.io/cpsievert/pen/MqLxOM

The `style()` function in R is meant to mimic `Plotly.restyle()`, but it doesn't currently support the same "partial update" API.

```r
p <- plot_ly(x = 1:10, y = 1:10, symbol = I(15))
# as expected, produces red circles
style(p, marker = list(color = "red"))
# should produce red squares
style(p, marker.color = "red")
```

cpsievert:


jcheng5:
I don't think this is quite right. The plotly.js semantics are that whatever object is found at the property path, is completely replaced. Your semantics appear to be more like if there is a `.` in the property path, then do a merge. Here's an example that demonstrates the difference:

```r
p <- plot_ly(x = 1:10, y = 1:10, symbol = I(15), marker = list(
  line = list(
    width = 5,
    color = "orange"
  )
))
p

# According to plotly.js semantics, this should make the marker line
# back to its default color, but with this PR it's still orange
style(p, marker.line = list(width = 10))
```

Issue #985, opened 2017-04-28
Ignoring explicitly provided widget ID; Shiny doesn't use them


komalsrathi:
I am using the current version of plotly and shiny. Whenever I am trying to create a plot, I am getting a warning:

```
Ignoring explicitly provided widget ID "fcfd29930da9"; Shiny doesn't use them
```

I have tested it with and without using plotly, and the warning seems to be associated with plotly.

Shiny and plotly versions:
```
plotly_4.6.0.9000
shiny_1.0.3
```

cpsievert:
It should be safe to ignore this warning. I decided to add an id to avoid [this warning](https://github.com/rstudio/rmarkdown/issues/980) in rmarkdown documents, which in mind, is more annoying

daattali:
@cpsievert is there any way to suppress this warning (without a global side effect like suppressing all warnings)?

cpsievert:
This hack does the trick for now, but they should really merge this PR https://github.com/ramnathv/htmlwidgets/pull/265

```r
library(shiny)
library(plotly)

ui <- fluidPage(
  plotlyOutput("p")
)

server <- function(input, output, ...) {
  
  output$p <- renderPlotly({
    p <- plotly_build(qplot(1:10))
    p$elementId <- NULL
    p
  })
  
}

shinyApp(ui, server)
```

daattali:
Awesome, as long as you don't think `p$elementId <- NULL` might breaks in a future date, I'm happy to use that :)

simonsays1980:
While the warning has vanished with setting `elementId <-  NULL`, it produces probably another warning when stopping the app: 

`Warning message:
In rm(list = as.character(handle), pos = .wsconns) :
  object '4516200352' not found`

vamsitroy:
hi,
using the above fix, I am still unable to generate a plot, a plot background just pops up without any kind of plot on it. please Advise.

cpsievert:
@vamsitroy this workaround is no longer required with the dev version:

```r
devtools::install_github('ropensci/plotly')
```

emrichter:
@cpsievert I downloaded the lastest dev version of plotly. I'm trying to test my shiny-app using shinytest.
It finds a difference between expected and current, since the elementId is everytime newly generated. Is there a workaround?
Thank in advance!

Here the screenshot:
![grafik](https://user-images.githubusercontent.com/36728688/36532723-cd4003fe-17c1-11e8-8b0c-e7ce0a634114.png)

[...]

PR #1365, opened 2018-10-10
Avoid using tempfile() for ID

krlmlr:
This changes `new_id()` to use `runif()` and `digest::digest()` to generate reproducible IDs. This is important for testing and resolves https://github.com/rstudio/shinytest/issues/174 and https://github.com/ropensci/plotly/issues/985#issuecomment-367634260.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -88,7 +88,7 @@ deparse2 <- function(x) {
 }
 
 new_id <- function() {
-  basename(tempfile(""))
+  digest::digest(runif(3), "xxhash64")
 }
 
 names2 <- function(x) {


cpsievert:
Thanks for the PR @krlmlr! Although it's definitely a corner-case, this change would break cases where one has (probably inadvertently) reset the same seed in-between adding of new layers.

```r
set.seed(1)
p <- plot_ly()
set.seed(1)
p <- add_markers(p, data = mtcars[1:10, ], x = ~wt, y = ~mpg, color = I("black"))
set.seed(1)
add_paths(p, data = mtcars[11:20, ], x = ~wt, y = ~mpg, color = I("black"))
```

The reasons have to do with the way plotly works internally -- it uses `new_id()` to attach an ID to each graphical layer and it's currently not smart enough to deal with non-unique IDs.

I'll contact @wch and see if we can't add some special logic in shinytest to provide a better experience plotly+shinytest experience without such a fundamental change to plotly

wch:
FYI: it's possible to call `snapshotPreprocessOutput()` to add an output preprocessing function to clean up the raw output value to something that's more deterministic for testing. See:
https://github.com/rstudio/shiny/blob/6ede019/R/snapshot.R#L10-L19

For an example of usage, see: https://github.com/rstudio/DT/blob/80981d3/R/shiny.R#L116-L124

cpsievert:
Closing in favor of #1379

PR #1369, opened 2018-10-16
leverage custom writer in vdiffr

cpsievert:
These changes rely on https://github.com/lionel-/vdiffr/pull/48. That PR should be merged first, but strictly speaking, it doesn't have to since the dependency managed via https://hub.docker.com/r/cpsievert/plotly-orca/)


cpsievert:
Waiting on https://github.com/lionel-/vdiffr/pull/48

PR #1379, opened 2018-10-29
Improve plotly+shinytest experience

cpsievert:
This PR addresses the main problem in https://github.com/rstudio/shinytest/issues/174, but it's worth noting there are a couple remaining issues to be aware of:

* Since shinytest currently runs on phantomjs, it won't work well with plotly's WebGL chart types. a9714dd adds a warning if plotly detects webgl in shinytest mode, but until https://github.com/rstudio/shinytest/issues/146 is addressed, there is no real way for users to see that warning
* How should input values that are triggered through direct manipluation events be handled? I suppose a more long-term solution would be to leverage whatever comes out of https://github.com/rstudio/shinytest/issues/193, but a shorter-term fix might be to use `shiny::snapshotPreprocessInput()`?

## Testing notes

Install `devtools::install_github("ropensci/plotly#1379")`, then use **shinytest** to record and test the following app -- https://github.com/ropensci/plotly/tree/master/inst/examples/shiny/Diamonds

Make sure you don't get false positive test failures when running the tests.

Here is some code to help with those tasks:

```shell
git clone https://github.com/ropensci/plotly.git
cd plotly
R
shinytest::recordTest("inst/examples/shiny/Movies/")
shinytest::testApp("inst/examples/shiny/Movies/")
```


R/shiny.R
```diff
@@ -39,7 +39,18 @@ renderPlotly <- function(expr, env = parent.frame(), quoted = FALSE) {
...
-  shinyRenderWidget(expr, plotlyOutput, env, quoted = TRUE)
+  renderFunc <- shinyRenderWidget(expr, plotlyOutput, env, quoted = TRUE)
+  # remove 'internal' plotly attributes that are known to cause false
+  # positive test results in shinytest (snapshotPreprocessOutput was added 
+  # in shiny 1.0.3.9002, but we require >= 1.1)
+  shiny::snapshotPreprocessOutput(
+    renderFunc,
+    function(value) {
+      json <- from_JSON_safe(value)
+      json$x <- json$x[!names(json$x) %in% c("visdat", "cur_data", "attrs")]
```
wch: This isn't important, but I find it slightly clearer to use `setdiff()`, as in:

```R
 json$x[setdiff(names(json$x), c("visdat", "cur_data", "attrs"))]
```

PR #1382, opened 2018-10-31
use hoverinfo skip instead of none for invisible ggplotly() colorbar trace, closes #1381

### LINKED ISSUE #1381: ggplotly()'s colorbar trace can override hover behavior
Thanks to @goodekat for the reprex

```r
library(plotly)
library(tidyr)

m <- matrix(rnorm(9), nrow = 3, ncol = 3)

m_gathered <- data.frame(m) %>%
  gather(key = column) %>%
  mutate(row = factor(rep(c("X1", "X2", "X3"), 3))) %>%
  select(column, row, value)

p <- ggplot(m_gathered, aes(x = column, y = row, fill = value)) +
  geom_tile()

# Apply plotly to ggplot heatmap - deadspace in the middle of (X1, X1)
ggplotly(p)
```

hover

The fix seems easy -- change `hoverinfo="none"` to `hoverinfo="skip"` when creating the colorbar trace

https://github.com/ropensci/plotly/blob/7031a25474f0dc275d20897810682c530cd1fbf3/R/ggplotly.R#L1303

cpsievert:
See #1381

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -1300,7 +1300,7 @@ gdef2trace <- function(gdef, theme, gglayout) {
       type = "scatter",
       mode = "markers",
       opacity = 0,
-      hoverinfo = "none",
+      hoverinfo = "skip",
       showlegend = FALSE,
       # do everything on a 0-1 scale
       marker = list(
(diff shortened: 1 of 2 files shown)


schloerke:
So the grid lines were taking the hover info first, even though they are underneath the tile?

PR #1388, opened 2018-11-01
Fix the definition of has_fill()

cpsievert:
## Background

After the `stroke` argument was recently introduced (v4.8.0 I believe), the `color` argument was refined to control the "fill color" properties (previously, it would control both "stroke color" and "fill color"). 

That logic relies on `has_fill()` to determine whether "fill color" is relevant for the trace (if it isn't, then both `color` and `stroke` will control the "stroke color")

This PR broadens the definition of what it means for a plotly trace to have a "fill color" to include traces that have a `fillcolor` attribute but no `fill` attribute (e.g., `box` traces).

## Testing notes

Please ensure both of the examples below produce the following image:

screen shot 2018-11-05 at 3 38 20 pm

```r
# devtools::install_github("ropensci/plotly#1388")
library(plotly)

set.seed(1)
data <- data.frame(x = rep(c("A", "B"), each = 5), y = rnorm(10))
cols <- c('A' = "blue", 'B' = "red")

plot_ly(data) %>% 
  add_boxplot(x = ~x, y = ~y, color = ~x, colors = cols, line = list(color = "black"))
```

```r
plot_ly(data) %>% 
  add_boxplot(x = ~x, y = ~y, color = ~x, colors = cols, stroke = I("black"))
```


jcheng5:
I've looked at this a couple of times and can't understand what's happening. If you're comfortable merging it, that's fine with me, but if you feel like you need useful feedback from me it's going to require a top-down tour first.

My only tactical question is that I would've expected a unit test for `has_fill` specifically, given how many conditionals and defaults are present.

PR #1392, opened 2018-11-02
Add more input events

cpsievert:
See changes to NEWS.md for the relevant improvements.

## Testing Notes

Install `devtools::install_github("ropensci/plotly#1392")`. Run any apps under https://github.com/ropensci/plotly/tree/input-events/inst/examples/shiny that start with 'event_data' and verify they do a sensible thing (happy to walk you through that)


cpsievert:
Note to self: This example was working as of 6e9e636, but recent commits made (i.e., the new send message approach brakes it)

```r
library(shiny)
library(dplyr)
library(nycflights13)
# install from https://github.com/hadley/ggstat
library(ggstat)

# Collect numeric variables to be shown as histograms in the crossfilter
d <- select(flights, arr_time, dep_time, arr_delay, dep_delay, air_time, distance)

# Generate an output container per variable
ui <- fluidPage(
  lapply(names(d), function(nm) plotlyOutput(nm, height = 200, width = "49%", inline = TRUE))
)

server <- function(input, output, session) {
  
  # These reactive values track the set of active brushes 
  # Each reactive value corresponds to a different variable 
  brush_ranges <- reactiveValues()
  
  # Function to filter the dataset based on every active brush range
  # except for one -- this is needed since each view respect every brush 
  # except for it's own
  d_filter <- function(d, var = "arr_time") {
    for (nm in setdiff(names(d), var)) {
      rng <- brush_ranges[[nm]]
      if (is.null(rng)) next
      d <- filter(d, between(d[[nm]], min(rng), max(rng)))
    }
    d
  }
  
  # Implement same render logic for each variable
  lapply(names(d), function(nm) {
    
    counts <- d[[nm]] %>%
      bin_fixed(bins = 150) %>%
      compute_stat(d[[nm]]) %>%
      filter(!is.na(xmin_)) %>%
      mutate(xmid = (xmin_ + xmax_) / 2)
    
    output[[nm]] <- renderPlotly({
      
      plot_ly(counts, source = nm) %>%
        add_bars(x = ~xmid, y = ~count_) %>%
        layout(
          dragmode = "select", 
          selectdirection = "h",
          xaxis = list(
            title = nm,
            range = range(d[[nm]], na.rm = TRUE)
          ),
          yaxis = list(title = "")
        )
    })
    
    observeEvent(event_data("plotly_brushing", source = nm), ignoreNULL = FALSE, {
      
      # inform the world about the new brush range
      brush_ranges[[nm]] <- event_data("plotly_brushing", source = nm)$x
      
      # update the bar heights of every view (except for the one being brushed)
      for (var in setdiff(names(d), nm)) {
        
        # views respect every brush except for their own 
        d_filtered <- d_filter(d, var)
        
        # bin the filtered data based on the global binning definition
        counts_filter <- d[[var]] %>%
          bin_fixed(bins = 150) %>%
          compute_stat(d_filtered[[var]]) %>%
          filter(!is.na(xmin_)) %>%
          mutate(xmid = (xmin_ + xmax_) / 2)
        
        plotlyProxy(var, session) %>%
          plotlyProxyInvoke("restyle", "y", list(counts_filter$count_), 0) 
      }
    })
    
  })
  
  
  
}

shinyApp(ui, server)
```

jcheng5:
Example where the hover data doesn't seem to update correctly:

```r
library(shiny)
library(plotly)

ui <- fluidPage(
  plotlyOutput("plot")
)

server <- function(input, output, session) {
  output$plot <- renderPlotly({
    plot_ly(economics, x = ~pop, y = ~unemploy)
  })
  
  observe({
    print(event_data("plotly_hover"))
  })
}

shinyApp(ui, server)
```

inst/htmlwidgets/plotly.js
```diff
@@ -893,3 +820,140 @@ function debounce(func, wait, immediate) {
...
+      plotly_deselect: ["plotly_selected", "plotly_selecting", "plotly_brushed", "plotly_brushing", "plotly_click"],
+      plotly_unhover: ["plotly_hover"],
+      plotly_doubleclick: ["plotly_click"]
+    }
+        
+    // register events for all DOM elements connected to this source id
+    var idMap = crosstalk.var("plotlySourceDomMap").get() || {};
+    var gids = idMap[src];
+    gids.map(function(id) {
+      var gid = document.getElementById(id);
```
jcheng5: Make sure `gid` is truthy, element could've been removed

jcheng5:
If you brush the plot then hit the button, you get `NULL` instead of the brush event data.

```r
library(shiny)
library(plotly)

ui <- fluidPage(
  plotlyOutput("plot"),
  actionButton("show_info", "Dump some info")
)

server <- function(input, output, session) {
  output$plot <- renderPlotly({
    plot_ly(economics, x = ~pop, y = ~unemploy)
  })
  
  observeEvent(input$show_info, {
    print(event_data("plotly_brushed"))
  })
}

shinyApp(ui, server)
```

jcheng5:
Notes from discussion (not intended to be implemented):

```
1. DOM element that raises events
2. Events get serialized into input values
3. Events get associated with input names (".clientData-plotly_selected-A-event")
4. Events may or may not have been requested by the server
5. If so, events are sent to server

renderValue: function(x) {
  el.on("plotly_brushed", evt => {
    let event_id = "plotly_brushed-" + el.source + "-event";
    let value = eventDataWithKey(evt);
    window.most_recent_plotly_event_values[event_id] = value;
    if (window.is_plotly_event_registered[event_id]) {
      Shiny.setInputValue(...)
    }
  });
}


Shiny.addCustomMessageHandler("plotly_event", msg => {
  if (window.is_plotly_event_registered[msg.id]) return;
  
  window.is_plotly_event_registered[msg.id] = true;
  Shiny.setInputValue(msg.id, window.most_recent_...);
});


Data structures in client
- A list of event/src/priority that have been requested by server
- The most recent values of all event/src/priority

|----------------|   [ ^ Show info          ]
|    x           |
|      x         |
|    x    x      |
|----------------|
```

R/shiny.R
```diff
@@ -115,13 +115,20 @@ event_data <- function(
...
+    # event-source combo is legitimate. As a result, we throw a 
+    # warning here to try to be useful in cases where you really
+    # are requesting an input value that is never registered, but
+    # without preventing valid code from running.
+    warning(
+      "The '", event, "' event tied a source ID of '", source, "' ",
+      "has not yet been registered. ", 
+      "If you encounter issues acquiring data from this event-source combo",
+      "Try adding `event_register(p, '", event, "')` to plot `p`"
     )
```
cpsievert: This is gonna lead to a lot of false positive warnings (like in the app below), but I see no other great way around it. 

```r
library(shiny)

ui <- fluidPage(
    plotlyOutput("p")
)

server <- function(input, output, session) {
    
    observe({
        print(event_data("plotly_click"))
    })
    
    output$p <- renderPlotly({
        plot_ly(mtcars, x = ~mpg, y = ~wt)
    })
    
}

shinyApp(ui, server)
```

jcheng5: Why's that--because the observer executes first?

jcheng5: If that's the issue, you could defer the check/warning using `session$onFlushed`. That'll ensure that anybody who was going to register those events in the same "tick" will have done so.

cpsievert: Oh, yes, good idea!

R/shiny.R
```diff
@@ -116,21 +116,26 @@ event_data <- function(
...
+  # It's possible for event_data() to execute before any 
+  # relevant input values have been registered (i.e, before 
+  # relevant plotly graphs have been executed). Therefore, 
+  # we delay checking that a relevant input value has been 
+  # registered until shiny flushes
+  session$onFlushed(
+    function() {
+      eventIDRegistered <- eventID %in% session$userData$plotlyShinyEventIDs
+      if (!eventIDRegistered) {
+        stop(
```
jcheng5: I'd warn here, not stop. I'm not sure if throwing an error here interrupts other `onFlushed`, but even if not, I think it's closer to the intent if we just warn here, since there's no code on the stack that could even do anything sensible in response to this error, other than log it.

PR #1407, opened 2018-11-15
Properly retain original opacity of 0

### LINKED ISSUE #1406: Plotly generates artificial data point when using with crosstalk
When using with crosstalk, plotly generates "new" data point on the scatter plot after selecting and deselecting.

First plot a scatter plot with plotly and crosstalk.
image

Then highlight some points by selecting them.
image

Double click anywhere to deselect. Then these's a "new" data point shows at the lower left corner.
image

If I set `color = I("black")`, everything works fine.

R 3.4.1, plotly 4.8.0, crosstalk 1.0.0, macOS High Sierra 10.13.6.

```r
library(plotly)
library(crosstalk)
library(tidyverse)

m <- mtcars %>% tibble::rownames_to_column()
d <- SharedData$new(m, ~rowname)

plot_ly(d, x = ~mpg, y = ~disp, color = ~cyl, type="scatter", mode = "markers") %>%
    highlight(on="plotly_selected", off="plotly_deselect")
```

cpsievert:
Closes #1406. 

The bug in #1406 happens because of the hidden trace that is used to render the colorbar. That trace has an opacity of 0, but `0 || 1` evaluates to 1 in JavaScript, so this client-side logic used to store the original opacity (needed for 'clearing' a crosstalk event) was incorrectly storing 1 (instead of 0), making that invisible trace visible when clearing the brush.

## Testing notes

Install (`devtools::install_github("ropensci/plotly#1407")`) then make sure the behavior outlined in #1406 is fixed (i.e. a marker in the lower-left should not appear after click&drag over points and double-clicking to clear the highlight)

### DIFF
--- a/inst/htmlwidgets/plotly.js
+++ b/inst/htmlwidgets/plotly.js
@@ -522,7 +522,7 @@ function TraceManager(graphDiv, highlight) {
   // avoid doing this over and over
   this.origOpacity = [];
   for (var i = 0; i < this.origData.length; i++) {
-    this.origOpacity[i] = this.origData[i].opacity || 1;
+    this.origOpacity[i] = this.origData[i].opacity === 0 ? 0 : (this.origData[i].opacity || 1);
   }
 
   // key: group name, value: null or array of keys representing the


schloerke:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators

> However, the && and || operators actually return the value of one of the specified operands, so if these operators are used with non-Boolean values, they may return a non-Boolean value.

Issue #1401, opened 2018-11-13
Using Shift on event_data("plotly_click") only stores last key


trafficonese:
When using the SHIFT-key on a `plotly_click` event, it seems to only store the last clicked key, whereas using SHIFT on `plotly_selected` stores all selected keys. 

This seems weird, as the highlighting shows all clicked elements. 

Is it possible to adapt the `event_data("plotly_click")` so that it corresponds with the highlighting and returns all clicked elements?
Or can I deactivate SHIFT for click-events alltogether and leave it activated for selection-events?

![clickselect](https://user-images.githubusercontent.com/35805180/48405274-4661dd80-e732-11e8-9aba-328a2d1620c0.png)

Shiny-App:
```
library(shiny)
library(ggplot2)
library(plotly)

dfN <- data.frame(
  time_stamp = seq.Date(as.Date("2018-04-01"), as.Date("2018-07-30"), 1),
  val = runif(121, 100,1000),
  col = "green", stringsAsFactors = F
)


ui <- fluidPage(
  plotlyOutput("plot"),
  verbatimTextOutput("clicked"),
  verbatimTextOutput("selection")
)

server <- function(input, output, session) {
  output$plot <- renderPlotly({
    key <- highlight_key(dfN)
    p <- ggplot() +
      geom_col(data = key, aes(x = plotly:::to_milliseconds(time_stamp), y = val, fill=I(col))) +
      theme(legend.position="none")
    
    ggplotly(p, source = "Src") %>% layout(xaxis = list(tickval = NULL, ticktext = NULL, type = "date")) %>% 
      highlight(selectize=F, off = "plotly_doubleclick", on = "plotly_click", color = "blue",
                opacityDim = 0.5, selected = attrs_selected(opacity = 1))
  })
    

  output$clicked <- renderPrint({
    s <- event_data("plotly_click", source = "Src")
    s
  })
  output$selection <- renderPrint({
    s <- event_data("plotly_selected", source = "Src")
    s
  })
}

shinyApp(ui, server)
```

cpsievert:
You raise a great point with this example and I don't think there currently is a great way adapt 'plotly_click' to accumulate data by holding shift. I will point out though that `event_data()` is designed to work both with and without `highlight_key()`, so it tries to not be too opinionated about what you want to do with it.

[...]

PR #1409, opened 2018-11-15
Access persistent click/hover data in shiny, addresses #1401

cpsievert:
See #1401 for more context

## Review questions

  - [ ] Is it safe to remember event data by attaching an attribute to `graphDiv` in this way? Should try this out in conjunction with `plotlyProxy()`.
  - [ ] Should we worry about emitting redundant data in the persistent case? Or should this be the responsibility of the user?
  - [ ] Should we worry that the persistent hover data isn't cleared when unhovering after a shift?

## Testing notes

Install (`devtools::install_github("ropensci/plotly#1409")`), run this app, and make sure the event data is what you'd expect.

```r
library(shiny)
library(plotly)

ui <- fluidPage(
  plotlyOutput("plot"),
  verbatimTextOutput("click"),
  verbatimTextOutput("shift_click"),
  verbatimTextOutput("hover"),
  verbatimTextOutput("shift_hover")
)

server <- function(input, output, session) {
  output$plot <- renderPlotly({
    plot_ly(x = 1:10, y = 1:10, key = LETTERS[1:10])
  })
  
  output$click <- renderPrint({
    d <- event_data("plotly_click")
    if (is.null(d)) "Transient click data" else d
  })
  output$shift_click <- renderPrint({
    d <- event_data("plotly_click_persist_on_shift")
    if (is.null(d)) "Persistent click data" else d
  })
  output$hover <- renderPrint({
    d <- event_data("plotly_hover")
    if (is.null(d)) "Transient hover data" else d
  })
  output$shift_hover <- renderPrint({
    d <- event_data("plotly_hover_persist_on_shift")
    if (is.null(d)) "Persistent hover data" else d
  })
}

shinyApp(ui, server)
```

### DIFF
--- a/inst/htmlwidgets/plotly.js
+++ b/inst/htmlwidgets/plotly.js
@@ -277,29 +277,71 @@ HTMLWidgets.widget({
         );
       });
       graphDiv.on('plotly_hover', function(d) {
+        
         Shiny.onInputChange(
           ".clientValue-plotly_hover-" + x.source, 
           JSON.stringify(eventDataWithKey(d))
         );
-      });
-      graphDiv.on('plotly_click', function(d) {
+        
+        // When persistent selection is enabled (either through
+        // shift or highlight()), remember click data
+        if (x.highlight.persistentShift) {
+          var dShift = graphDiv._shiny_plotly_hover || {points: []};
+          var pts = [].concat(dShift.points, d.points);
+          var d = {points: pts, event: d.event};
+          graphDiv._shiny_plotly_hover = d;
+        } else {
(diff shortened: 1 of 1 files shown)


trafficonese:
Shouldn't a doubleclick clear the Shift selections? 
Now the selection is only cleared, by clicking a new point/bar.

It also doesnt always reflect the highlighted elements, as Shift must be pressed already when clicking the first element. The "usual" behaviour would be to click the first without Shift and then to add more into the selection, hold Shift and click another point which will be added to the current selection. 

The Box/Lasso-selection behave like the "usual" way.

Otherwise nice work 👍

PR #1450, opened 2019-01-09
upgrade plotly.js

cpsievert:
## TODO

These plotly.js release notes will require some R changes:

* `title` attributes linked to strings are now deprecated. Please use `title.text` instead to fill in your title text [#3276]
* `title*` attributes are now deprecated. They moved to `title.*`. For example, colorbar.titleside is now colorbar.title.side [#3276]

FWIW, I'm pretty sure `layout.title`, `layout.titlefont`, `layout[axisid].title`, `layout[axisid].titlefont`, `marker.colorbar.title`, `marker.colorbar.titlefont`, `marker.colorbar.titleside`, and [carpet-specific attributes](https://github.com/plotly/plotly.js/pull/3276/files#diff-966c586836ba9876294bcca2af38b5eb)  are the only ones effected. 

- [x] It looks like these attributes still work, but be good form to identify these cases on the R side and throw an information deprecation message/warning. 
- [ ] Update any 'internal' logic using these attributes to use the new nested structure.
- [ ] Would also be a good time to have `ggplotly()` leverage the new `layout.title` and `layout[axisid].title` placement attributes


cpsievert:
Closing in favor of #1506

PR #1477, opened 2019-03-06
sec_axis support

adrianoesch:
```
ggplotly(ggplot(mtcars,aes(x=drat,y=hp))+geom_point()+
  geom_point(aes(y=qsec,color='red',axis=2))+
  scale_y_continuous(sec.axis = sec_axis(~./6,'2nd Axis')))
```

an additional `axis='sec'` or  `axis=2` is necessary in the geom for ggplotly to know which trace needs to be mapped onto y2, which results in a warning from ggplot (`Ignoring unknown aesthetics: axis`) but the outcome seems good enough to me. let me know what you think (tests included).

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -562,9 +562,20 @@ gg2list <- function(p, width = NULL, height = NULL,
   )
   doms <- get_domains(nPanels, nRows, margins)
   
+  # check if secondary axis
+  y_attr = attributes(scale_y()$secondary.axis)
+  if('AxisSecondary' %in% y_attr$class){
+    layout_axes = c('x','y','y2')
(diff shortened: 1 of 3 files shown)


cpsievert:
Thanks for your effort on this @adrianoesch, I appreciate it.

I'm sorry to say though that I think the complexity that secondary axes will introduce in `ggplotly()` may not be worth the trouble. Ideally, this should work for facets (and without extending ggplot2), which will likely increase the global complexity of `ggplotly()`. Perhaps I'll change my mind if I get enough time to overhaul things, but I just want to make it clear this feature is not very likely to come from a community contribution because there are so many other factors involved.

Issue #1493, opened 2019-03-18
subplot() doesn't handle xsizemode/ysizemode properly


cpsievert:
I'd expect this code example to create facet strips with a fixed height for each panel, but it currently doesn't:

```r
library(plotly)
library(dplyr)

my_plot <- . %>% 
  plot_ly(x = ~date, y = ~value) %>%
  add_annotations(
    text = ~unique(variable),
    x = 0.5,
    y = 1,
    yref = "paper",
    xref = "paper",
    xanchor = "middle",
    yanchor = "bottom",
    showarrow = FALSE,
    font = list(size = 15)
  ) %>%
  layout(
    shapes = list(
      type = "rect",
      x0 = 0,
      x1 = 1,
      xref = "paper",
      y0 = 0, 
      y1 = 16,
      yanchor = 1,
      yref = "paper",
      ysizemode = "pixel",
      fillcolor = toRGB("gray80"),
      line = list(color = "transparent")
    )
  )

economics_long %>%
  group_by(variable) %>%
  do(p = my_plot(.)) %>%
  subplot(nrows = NROW(.), shareX = TRUE)
```

Issue #1333, opened 2018-09-05
Shapes/images don't accumulate with multiple calls to layout()


cpsievert:
```r
line1 <- list(
  type = "line",
  x0 = 0,
  x1 = 1,
  y0 = 0.5,
  y1 = 0.5,
  xref = "paper",
  yref = "paper",
  line = list(color = "#C3C4C6")
)

line2 <- list(
  type = "line",
  x0 = 0,
  x1 = 1,
  y0 = 0.75,
  y1 = 0.75,
  xref = "paper",
  yref = "paper",
  line = list(color = "#C3C4C6")
)

plotly_empty() %>%
  layout(shapes = line1) %>%
  layout(shapes = line2) 
```

cpsievert:
This also appears to be a problem when accumulating shapes in `subplot()`?

```r
library(dplyr)
library(plotly)
set.seed(123)

dat <- data.frame(
  a = rnorm(100), b = rnorm(100), c = rnorm(100),
  class = sample(c("g1", "g2", "g3"), 100, replace = TRUE),
  grp = sample(c("n", "o", "p", "q", "r"), 100, replace = TRUE))


plots <- dat %>%
  group_by(class) %>%
  do(plot = {
    plot_ly(., x = ~a, y = ~b, legendgroup = ~grp, 
            showlegend = .$class[1] == "g1") %>%
      add_markers(color = ~grp) %>%
      layout(
          xaxis = list(showzeroline = FALSE), 
          yaxis = list(showzeroline = FALSE),
          shapes = list(
              type = "line",
              x0 = 0,
              x1 = 0, 
              y0 = 0, 
              y1 = 1,
              xref = "paper",
              yref = "paper"
          ),
          list(
              type = "line",
              x0 = 0,
              x1 = 0, 
              y0 = 0, 
              y1 = 1,
              xref = "paper",
              yref = "paper"
          ))
  })
subplot(plots, nrows = 2, shareX = TRUE, shareY = TRUE)
```

Related https://community.rstudio.com/t/adding-borders-to-all-subplots-in-plotly-when-sharing-axes/30289

mehrnoushmalek:
@cpsievert Hi Carson, do you know if having multiple subplot shapes is fixed or will be fixed anytime soon?
I'd like to have two rectangle shape (one around the subplot title, and one in the plot itself), but I can only add one of them. This is the case for just one plot as well.
[...]

PR #1494, opened 2019-03-18
reposition [x-y]anchor when [x-y]sizemode='pixel'

cpsievert:
See #1493 for an example

## Background

`layout.shapes` have `xsizemode`/`ysizemode` attributes that allow you to specify a fixed height/width for shapes by setting, say `ysizemode='pixel'`. In this case, the `y0`/`y1` are measured in pixels and are positioned relative to `yanchor`. Therefore, when `ysizemode='pixel'` and `yref = 'paper'`, `subplot()` needs to reposition `yanchor` instead of `y0`/`y1`


tests/testthat/test-plotly-subplot.R
```diff
@@ -208,6 +208,56 @@ test_that("shape paper repositioning", {
...
+        x1 = 30,
+        xanchor = 0.5,
+        xsizemode = "pixel",
+        xref = "paper",
+        line = list(color = "blue")
+      )
+    )
+  
+  s <- subplot(p1, p2)
+  shapes <- expect_doppelganger_built(s, "subplot-reposition-shape-fixed")$layout$shapes
```
cpsievert: Add a rect shape after-the-fact as well

tests/testthat/test-plotly-subplot.R
```diff
@@ -208,6 +208,56 @@ test_that("shape paper repositioning", {
...
+        yanchor = 0.5,
+        ysizemode = "pixel",
+        yref = "paper",
+        fillcolor = "red"
+      )
+    )
+  p2 <- plot_ly(mtcars) %>%
+    layout(
+      shapes = ~list(
+        type = "rect",
```
cpsievert: remove formula

tests/testthat/test-plotly-subplot.R
cpsievert: I forgot that you can't currently add shapes to a plot that already has shapes -- https://github.com/ropensci/plotly/issues/1333

PR #1498, opened 2019-03-20
subplot(): avoid Reduce() when merging lists

alyst:
The PR replaces the uses of `Reduce(c, ...)` to merge the lists with `unlist(..., recursive = FALSE)`.

The motivation: for the list `x` of *n* elements `Reduce(c, x)` would implicitly create *n* lists/vectors of lengths 1 to *n*, whereas only the last one is needed. The pattern adds a considerable overhead when the subplots contain a lot of data.

It also simplifies `ensure_one()`: instead of using `[` to get the one-element list, it uses `[[` to get this one element (nulls are automatically filtered out).


cpsievert:
Thanks @alyst , this looks great! Do you have an example in mind where db622ad is relevant?

R/subplots.R
```diff
@@ -290,8 +291,8 @@ subplot <- function(..., nrows = 1, widths = NULL, heights = NULL, margin = 0.02
     }
     layouts <- layouts[which_layout]
   }
-  p$attrs <- Reduce(c, lapply(plots, "[[", "attrs"))
-  p$layout <- modify_list(p$layout, Reduce(modify_list, layouts))
+  p$attrs <- unlist(lapply(plots, "[[", "attrs"), recursive = FALSE)
+  p$layout <- Reduce(modify_list, layouts, p$layout)
```
jcheng5: If there are a lot of layouts then `do.call(modify_list, c(list(p$layout), layouts))` might be better (or something like that with judicious `quote()` calls added--@wch can give you details if you choose to go the `do.call` route). If there aren't a lot of layouts then this is fine with me.

PR #1501, opened 2019-03-20
ggplotly(): fix Xaxis anchor if the last row incomplete

alyst:
This is the continuation of the closed PR #619 after rebasing the changes to the current master.

*ggplotly* handles the `facet_wrap()` with the incomplete last row differently from *ggplot2*: `ggplot()` attaches the X axis to the last plot in the column, whereas `ggplotly(ggplot())` always puts it to the last row, even when there are no plots at that position. IMHO the *ggplot2* behaviour is better -- there are no "floating" axes; the axes are always attached to the plots.

There's one testcase affected by the fix: *"facet_wrap-free_y-2"* (it also demonstrates the difference between *ggplot* and current *ggplotly* behaviours).
I guess I also need to fix the reference *.png* image, but I'm not sure how to properly do it.


R/ggplotly.R
```diff
@@ -459,27 +459,35 @@ gg2list <- function(p, width = NULL, height = NULL,
...
+      layout$layout <- dplyr::mutate(layout$layout,
+        yaxis = PANEL,
+        yanchor = COL)
     }
+    # anchor X axis to the lowest plot in its column
+    layout$layout <- dplyr::group_by(layout$layout, xaxis) %>%
+      dplyr::mutate(xanchor = max(as.integer(yaxis))) %>%
+      dplyr::ungroup() %>%
+      dplyr::mutate(xanchor = if (is.factor(yaxis)) levels(yaxis)[xanchor] else xanchor)
   }
```
cpsievert: I don't immediately follow why `dplyr::mutate(xanchor = if (is.factor(yaxis)) levels(yaxis)[xanchor] else xanchor)` is necessary, do you have an example in mind where this becomes relevant?

cpsievert:
Thanks @alyst, again looking great! It'd be really awesome if you could either point out an existing test or add another test for the code I mentioned in https://github.com/ropensci/plotly/pull/1501/files#r271834045

R/ggplotly.R
alyst: Yes, it looks a bit ugly. I had trouble with some of the plots (maybe even the ones in the tests) when either a PANEL or COL is a factor.
There was some error happening afterwards, but unfortunately I forgot the details. So I had to convert `xanchor` back to a factor.

cpsievert: @alyst tests are passing without it -- I would be very grateful if you can report a case where 3c8505230aca09c4adf4672f91ffc1e1cc4a32f0 doesn't do what you'd expect!

Issue #1484, opened 2019-03-10
Grouped boxplot data not centred on xaxis ticks, looses boxgap argument, and wont resize with subplots


l-jaye:
Hi @cpsievert! 

Here after your Shiny webinar, great stuff!

I'm having issues with the boxplot data aligning with x-axis tick marks in subplots. Please see below:

```r
library (data.table)
library (stringr)
library (plotly)
# some setup:
get_plot_title = function(this.plot.title, y.pos = 1, y.shift = 0){
  list(text = sprintf ("<b>%s</b>", this.plot.title), yshift = y.shift
       , xref = "paper", yref = "paper", yanchor = "center", xanchor = "center", align = "center"
       , x = 0.5, y = y.pos, showarrow = FALSE, font = list (size = 16, family = "Arial")
  )} 

generate_panel = function(dat, trace.name){
  my.plot = 
    plot_ly(dat, type = "box"
            , x= ~get(x_var), y= ~get(y_var), color = ~get(color_var)
            , whiskerwidth = 0.1
            , boxpoints = F
    )   %>% 
    layout(boxmode = "group", boxgroupgap = 0.1, boxgap = 0.15
           , plot_bgcolor = "#F5F5F5"
           , annotations = get_plot_title(trace.name, y.shift = 20)
           , xaxis = list(title = "", tickfont = list(size = 14, family = "Arial")
                          , tickmode = "linear", ticks = "outside")
           , yaxis = list(title = sprintf("<b>%s</b>", y_var), titlefont = list (size = 16, family = "Arial"))
           ) 
  my.plot #return
}

x_var = "cut"
y_var = "price"
color_var = "color" 
X_panel_var = "clarity"
test.dat = ggplot2::diamonds %>% setDT
```
### Without subplots, everything looks great:
```r
generate_panel(test.dat, trace.name = "all")
```
![image](https://user-images.githubusercontent.com/32018129/54088600-17512680-4336-11e9-9cdc-ec0cee3ba31f.png)

### Adding two horizontal subplots:
```r
test.dat[, get(X_panel_var) %>% levels][1:2] %>% 
  lapply(function(this.panel.var){
    generate_panel(dat = test.dat[this.panel.var, on = X_panel_var], trace.name = this.panel.var)
  }) %>% 
  subplot(shareY = T, shareX = F, titleX = F, nrows = 1, which_layout = 1, margin = 0.01)
```
[...]

Issue #1241, opened 2018-04-19
Add option "notch" (and "notchwidth") to geom_boxplot


januz:
I would like to add notches to boxplots like it is possible in ggplot. The feature has recently been implemented for the plotly API (https://community.plot.ly/t/add-option-notch-to-box-plots/7396/3) 

```r
library(plotly)

set.seed(1234)
dat <- data.frame(cond = factor(rep(c("A","B"), each=200)), rating = c(rnorm(200),rnorm(200, mean=.8)))

p <- ggplot(dat, aes(x=cond, y=rating)) + geom_boxplot(notch = TRUE)

# ggplot shows notch
p

# ggplotly does not
ggplotly(p)
```

Issue #1114, opened 2017-09-08
Boxplot outliers are shown in black using ggplotly


dayaefrail:
If I create a boxplot in ggplot2 and convert it using ggplotly command, the outliers are outlined in black.
Here is a simple example:
```
library(ggplot2)
library(plotly)

p <- ggplot(mpg, aes(class, hwy))
g <- p + geom_boxplot(aes(colour = "red"))
ggplotly(g)
```

ggplot would show this chart:
![image](https://user-images.githubusercontent.com/20057104/30224249-9db15b58-9493-11e7-8221-3f35f68ac12e.png)

whereas plotly would show this chart:
![image](https://user-images.githubusercontent.com/20057104/30224286-c93aac34-9493-11e7-9b11-80a8ff9490ff.png)

Is this something that can be fixed?

jonocarroll:
This persists even when the outliers should be discarded, [in the examples also](https://plot.ly/ggplot2/box-plots/#outliers)

```r
library(plotly)
set.seed(123)

df <- diamonds[sample(1:nrow(diamonds), size = 1000),]

p <- ggplot(df, aes(cut, price, fill = cut)) + 
  geom_boxplot(outlier.shape = NA) + 
  ggtitle("Ignore outliers in ggplot2")

# Need to modify the plotly object and make outlier points have opacity equal to 0
p <- plotly_build(p)

p$data <- lapply(p$data, FUN = function(x){
  x$marker = list(opacity = 0)
  return(x)
})

p
```

![screen shot 2017-11-17 at 6 54 06 pm](https://user-images.githubusercontent.com/9496865/32937494-baab4626-cbc8-11e7-8660-f4a85ed7254b.jpg)

petehilljnr:
I managed to set the opacity property of the outliers using the following code:

```{r}
library(ggplot2)
library(plotly)

set.seed(123)

df <- diamonds[sample(1:nrow(diamonds), size = 1000),]

p <- ggplot(df, aes(cut, price, fill = cut)) + 
  geom_boxplot(outlier.shape = NA) + 
  ggtitle("Ignore outliers in ggplot2")

# Need to modify the plotly object and make outlier points have opacity equal to 0
p <- plotly_build(p)

for(i in 1:length(p$x$data)) {
  p$x$data[[i]]$marker$opacity = 0
}

p
```
![image](https://user-images.githubusercontent.com/6835337/39455569-4db89fac-4d35-11e8-9f55-30eda7e2a9e9.png)

[...]

Issue #1098, opened 2017-08-10
Custom ggplot2 boxplot doesn't work ggplotly


doug-friedman:
When I create a ggplot2 boxplot with custom dimensions (see the bottom of http://ggplot2.tidyverse.org/reference/geom_boxplot.html#examples), plotly doesn't render the boxplot when I apply ggplotly. However, it works ok with a traditional boxplot where the dimensions are calculated by geom_boxplot/stat_boxplot.

```r
## Libraries
library(ggplot2)
library(gridExtra)
library(plotly)

## Using fake data with extreme values
fake_data = data.frame(
  y1 = c(rnorm(100, 2, 3), rnorm(10, 10, 15))
)


# ggplot2
g1 = ggplot(data=fake_data, aes(x ="cat1", y= y1)) + 
  #stat_boxplot(geom="errorbar") +
  geom_boxplot() +
  labs(title="Traditional Boxplot") +
  ylim(c(-5, 25))

g1

quants = quantile(fake_data$y1, c(0.05, .25, .5, .75, .95))
q_df = data.frame(t(quants))
names(q_df) = c("y5", "y25", "y50", "y75", "y95")


g2 = ggplot(data=q_df, aes(x ="cat1")) + 
  #stat_boxplot(geom="errorbar") +
  geom_boxplot(stat="identity", aes(, 
       ymin = y5, lower = y25, 
       middle = y50, upper = y75,
       ymax = y95)) +
  geom_point(data=subset(fake_data, (y1 > q_df$y95) | (y1 < q_df$y5)), aes(y=y1)) +
  labs(title="5%/95% Boxplot") +
  ylim(c(-5, 25))

g2



# plotly
plotly::ggplotly(g1)
plotly::ggplotly(g2) # DOESN'T WORK!

```

PR #1514, opened 2019-04-09
geom_boxplot() improvements

cpsievert:
This changes the default behaivor of ggplotly() to render `geom_boxplot()` as low-level polygons rather than the plotly.js trace type. As result, we'll be able to exactly replicate the ggplot2 result in many more cases. A list of relevant issues this PR should close: #1484, #1241, #1114, #1098, #866, #697, #706

TODO:

- [ ] Should we still support translation of `geom_boxplot()` to plotly.js box traces? If so, how?

### DIFF
--- a/R/layers2traces.R
+++ b/R/layers2traces.R
@@ -786,6 +786,8 @@ geom2trace.GeomBoxplot <- function(data, params, p) {
     frame = data[["frame"]],
     ids = data[["ids"]],
     type = "box",
+    notched = params[["notch"]],
+    notchwidth = params[["notchwidth"]],
     fillcolor = toRGB(
       aes2plotly(data, params, "fill"),
       aes2plotly(data, params, "alpha")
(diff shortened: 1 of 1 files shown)


cpsievert:
This should wait until we upgrade to plotly.js 1.52, then we can use the new 'pre-computed' feature of the box trace.

vcmorini:
Hello, would you know how we can apply these changes in the python library?
Thank you

dingying85:
image
Hi,
Regarding #697, it seems that problem is still there, when plot two subgroup boxplots side by side, if one group only have 1 subgroup data available, then the dots are not aligned.

cpsievert:
I also noticed that this throw an error when you consider #1603 

```r
> library(plotly)
> 
> plot <- ggplot(diamonds, aes(cut, price)) +
+   geom_boxplot() +
+   scale_y_continuous(trans = "sqrt", limits = c(0, 25000))
> 
> ggplotly(plot)
Error in mm2pixels(convert(u, "mm")) : Unit must be in millimeters
```

PR #1525, opened 2019-05-08
Fixed a typo

chsafouane:
"its" instead of "it's"


cpsievert:
Thanks! This project uses roxygen documentation, meaning the `man/` files aren't edited by hand, instead the documentation should be updated in `R/`... could you make the same change to `R/highlight.R`?

cpsievert:
I went ahead and did this in ecf3b39

PR #1599, opened 2019-08-15
Feature: convert ggplot to plotly by adding a layer

atusy:
By this PR, we can convert ggplot to plotly by implementing `gginteractive()`.
In addition, `gginteractive` has the `interactive` argument to control the conversion.

```r
ggplot(mtcars, aes(wt, mpg)) +
  geom_point() +
  gginteractive()
  # gginteractive(FALSE) # can suppress conversion
```

This is a sort of syntax sugar for `ggplotly()`, which has to wrap whole ggplot:

```r
ggplotly(
  ggplot(mtcars, aes(wt, mpg)) +
  geom_point()
)
```


cpsievert:
Interesting, thanks! Is there a way we could do this without introducing a new function? That is, could something like this work?

```r
ggplot(mtcars, aes(wt, mpg)) +
  geom_point() +
  ggplotly()
```

atusy:
Well, I wanted to do so.
I did not do it because the first argument of `ggplotly` is `last_plot()`.
If it is missing or `NULL`, we can do what we want by

```
if (missing(p)) return(ggintearactive())
# or
if (is.nuill(p)) return(gginteractive())
```

The only way is to add a new argument like

```
ggplotly <- function(.., layer = FALSE) {
  if (layer) return(gginteractive())
  ...
}
```

PR #1621, opened 2019-09-20
alter new_id() so the id depends on set.seed()

ThierryO:
The current implementation is not affected by `set.seed()`. Rerunning the same code results in different id's, even if `set.seed()` is set.

The new version is based on `sample()`, yielding stable ids in case the user sets `set.seed()`. The benefit is that rerunning the code with unchanged definition and data yields identical HTML code. This is useful when the output is stored under version control.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -93,7 +93,7 @@ deparse2 <- function(x) {
 }
 
 new_id <- function() {
-  basename(tempfile(""))
+  paste(sample(c(0:9, letters), 12, replace = TRUE), collapse = "")
 }
 
 names2 <- function(x) {


cpsievert:
Thanks @ThierryO. Just because I'm curious, why do need the id to respect `set.seed()`? It is possible to set the id of the widget element this way (if you need to refer to it in some JavaScript, for instance):

```
library(plotly)
library(htmlwidgets)
p <- plot_ly()
p$elementId <- "myId"
onRender(p, "function(el) { console.log(el.id); }") 
```

ThierryO:
That's OK for a single plot. It is harder when you are making a gitbook with lots of plots.

PR #1646, opened 2019-10-22
proposed change to avoid peeking at 'grid' unit internals

pmur002:
Hi

This is a suggested change for mm2pixels() and verifyUnit() to avoid them peeking at the internal representation of 'grid' units.  It makes use of a new function in R-devel called grid::unitType(), hence the R version conditioning.

Besides being a nicer way to do things, this change is important to help us with the transition to a new internal representation of 'grid' units for the next R release (https://github.com/thomasp85/grid/issues/14).

Paul


cpsievert:
Thank you Paul!

PR #1697, opened 2020-02-06
Legend position

cpsievert:
To be merged after #1695


cpsievert:
Oof, I just realized `legend` doesn't have a `len` attribute, which the algorithm in `plotly_guide_positions()` depends on. [Hopefully `len` comes to plotly.js relatively soon](https://github.com/plotly/plotly.js/issues/1340), so I'm deciding to shelf the legend positioning part of this PR for now

Issue #851, opened 2017-01-18
wrapping of labels not supported in ggplotly


earowang:
Hi Carson,

`ggplot2` works fine for the wrapping of long labels, while `ggplotly` doesn't support yet. A little example is as follows:

```r
library(plotly)
library(stringr)

df_x <- data.frame(
  x = "this is very loooooooooooong text to illustrate",
  y = 100
)

p <- ggplot(aes(x = x, y = y), data = df_x) +
  geom_bar(stat = "identity") +
  scale_x_discrete(labels = function(x) str_wrap(x, width = 10))
p
ggplotly(p)
```

Cheers,
Earo

cpsievert:
Thanks @earowang. f85a9dd adds support for translating R linebreaks to HTML

For now, you'll need to manually adjust margins to account for the multiple lines of text (`ggplotly()`isn't that smart, yet)

```r
ggplotly(p) %>% layout(margin = list(b = 100))
```

earowang:
Awesome! Thanks Carson.

PR #1700, opened 2020-02-11
Translate `\n` to `<br />` in fct labels, too

salim-b:
Before, this only worked for colums of type character. This PR adds support for columns of type factor.

The linebreak conversion was first introduced in [v4.6.0](https://github.com/ropensci/plotly/blob/master/NEWS.md#new-features--improvements-4), see also https://github.com/ropensci/plotly/issues/851


salim-b:
Sorry for the delay! I finally found some time to meet your requests.

> could you provide some examples of where this is relevant

This is relevant whenever you want to plot a data column of type factor with labels containing line breaks. Currently, the line breaks are only translated to HTML (`<br>`) when the column is of type character. But factor columns are not only very common but also very convenient to work with (using pkg forcats), i.e. to (re)arrange, extend, split into "buckets" etc.

### A very simple example with multiline factor labels

``` r
library(magrittr)

iris %>%
    dplyr::mutate(Species =
                      Species %>%
                      forcats::fct_relabel(paste0,
                                           "\n\n(third line)\n(fourth line)")) %>%
    plotly::plot_ly(x = ~Sepal.Length,
                    y = ~Species)
#> No trace type specified:
#>   Based on info supplied, a 'bar' trace seems appropriate.
#>   Read more about this trace type -> https://plot.ly/r/reference/#bar
```

grafik

<sup>Created on 2020-04-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### The same example when the column is converted to type character before plotting

``` r
library(magrittr)

iris %>%
  dplyr::mutate(Species =
                  Species %>%
                  forcats::fct_relabel(paste0,
                                       "\n\n(third line)\n(fourth line)") %>%
                  as.character()) %>%
  plotly::plot_ly(x = ~Sepal.Length,
                  y = ~Species)
#> No trace type specified:
#>   Based on info supplied, a 'bar' trace seems appropriate.
#>   Read more about this trace type -> https://plot.ly/r/reference/#bar
```
grafik

<sup>Created on 2020-04-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

> and also write a few tests?

I pushed https://github.com/ropensci/plotly/pull/1700/commits/841e7e9731b81b6d1d0f73182ac7951845220a41 which adds an additional test to `tests/testthat/test-plotly.R`. I'm very new to writing tests (or R packages to begin with), so please bear with me :)

cpsievert:
Thanks for the update, this is looking good @salim-b!

A couple minor nitpicks:

* Could you please remove the `%>%` and the **forcats** dependency in the tests?
* Please switch the order in the DESCRIPTION (I'd prefer that all single authors are listed first, then Plotly)

tests/testthat/test-plotly.R
```diff
@@ -317,3 +317,38 @@ test_that("toWebGL() shouldn't complain if it's already webgl", {
...
+
+test_that("Line breaks are properly translated (R -> HTML)", {
+  skip_if_not_installed(pkg = "forcats")
+  
+  # create target labels
+  suffix <- "\n\n(third line)\n(fourth line)"
+  
+  target_labels <- iris$Species %>%
+    unique() %>%
+    as.character() %>%
```
cpsievert, line 329: I think you can get rid of the `as.character()` here as `paste()` and `gsub()` will coerce to character

salim-b:
>* Could you please remove the `%>%` and the **forcats** dependency in the tests?

Sorry, I assumed it would be fine to use forcats since it's listed under `Suggests:` in `DESCRIPTION`. I have replaced it now with equivalent base R code.

Regarding `%>%`: The pipe operator is used throughout this test file. Why can't I use it, too?

>* Please switch the order in the DESCRIPTION (I'd prefer that all single authors are listed first, then Plotly)

Done :)

tests/testthat/test-plotly.R
salim-b: Ok, done.

salim-b:
@cpsievert I've added the missing news item in https://github.com/ropensci/plotly/pull/1700/commits/d60898a60fffc016c16f675a503c59d432a41d2a. I'd be glad if you could merge this PR. 🙂

cpsievert:
Thank you!

Issue #1310, opened 2018-07-25
orca() does not support file paths


emallickhossain:
I'm trying to export a static image in R using orca(), but it ignores the file path and just exports it to the current location of the script that I'm running. 

For example, if I run `orca(image, "/images/plot1.png")` from a script located in  `/code/script.R`, it saves it to `/code/plot1.png` instead of where I specified.

0navarro:
I have the same issue. Did you happen to find a workaround?

cpsievert:
Can also do `withr::with_dir("images", orca(img, "plot2.png"))`

PR #1727, opened 2020-03-27
Add dir argument to orca

wfjvdham:
Solves: https://github.com/ropensci/plotly/issues/1310

### DIFF
--- a/R/orca.R
+++ b/R/orca.R
@@ -11,6 +11,7 @@
 #' 
 #' @param p a plotly object.
 #' @param file output filename.
+#' @param dir sets the output directory. Defaults to current working directory.
 #' @param format the output format (png, jpeg, webp, svg, pdf, eps).
 #' @param scale Sets the image scale. Applies to all output images.
 #' @param width Sets the image width. If not set, defaults to `layout.width` value. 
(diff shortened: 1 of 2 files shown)


R/orca.R
```diff
@@ -84,6 +86,7 @@ orca <- function(p, file = "plot.png", format = tools::file_ext(file),
   
   args <- c(
     "graph", tmp, 
+    "-d", dir,
```
cpsievert, line 89:
```suggestion
"-d", normalizePath(dir, mustWork = TRUE),
```

R/orca.R
```diff
@@ -62,7 +63,8 @@
 #' }
 #' 
 
-orca <- function(p, file = "plot.png", format = tools::file_ext(file), 
+orca <- function(p, file = "plot.png", dir = getwd(),
```
cpsievert, line 66:
```suggestion
orca <- function(p, file = "plot.png", dir = ".", 
```

cpsievert:
Coming back to this now and realizing that it seems we can do without an additional `dir` argument (i.e., `file` could accept a relative file path)? I'm hesitant to add `dir` not only because it adds an additional arg, but also because it could break existing code

cpsievert:
Closing since `orca()` is now deprecated in favor of `save_image()`

Issue #1731, opened 2020-03-30
ggplotly error when xlab is element_blank()


doorisajar:
Reprex:

``` r
library(ggplot2)
library(plotly)

p <- ggplot(iris, aes(Petal.Length, Sepal.Width, color = Species)) + 
    geom_point() + 
    xlab(element_blank())

ggplotly(p)
#> Error in if (nchar(axisTitleText) > 0) {: argument is of length zero
```

That seems to be happening here:

https://github.com/ropensci/plotly/blob/ebb017a6fbd1a582787017f98e8e40cb4f1431b3/R/ggplotly.R#L817

`xlab("")` works, so the issue is definitely that `nchar` of a theme element that is set to `element_blank()` is `NULL`. The same check happens a few lines later:

https://github.com/ropensci/plotly/blob/ebb017a6fbd1a582787017f98e8e40cb4f1431b3/R/ggplotly.R#L839

I get the same error when setting `ylab` and `ggtitle` to `element_blank`, but not for setting the legend title to blank via `theme(legend.title = element_blank())`. 

I might be able to do a pretty minimal PR for this. The [tests](https://github.com/ropensci/plotly/blob/master/tests/testthat/test-ggplot-ggplotly.R) look reasonably straightforward.

doorisajar:
I submitted a PR for this: https://github.com/ropensci/plotly/pull/1732

PR #1732, opened 2020-03-30
Fix ggplotly error with NULL axis labs

doorisajar:
Resolves https://github.com/ropensci/plotly/issues/1731


R/ggplotly.R
```diff
@@ -814,7 +814,7 @@ gg2list <- function(p, width = NULL, height = NULL,
           bbox(axisTickText, axisObj$tickangle, axisObj$tickfont$size)[[type]] +
           bbox(axisTitleText, axisTitle$angle, unitConvert(axisTitle, "pixels", type))[[type]]
         
-        if (nchar(axisTitleText) > 0) {
+        if (!is.null(axisTitleText) && nchar(axisTitleText) > 0) {
```
cpsievert:
```suggestion
        if (nzchar(axisTitleText %||% "")) {
```

R/ggplotly.R
```diff
@@ -836,7 +836,7 @@ gg2list <- function(p, width = NULL, height = NULL,
           }
           # facets have multiple axis objects, but only one title for the plot,
           # so we empty the titles and try to draw the title as an annotation
-          if (nchar(axisTitleText) > 0) {
+          if (!is.null(axisTitleText) && nchar(axisTitleText) > 0) {
```
cpsievert:
```suggestion
          if (nzchar(axisTitleText %||% "")) {
```

R/ggplotly.R
doorisajar: Sure, let me just make sure that behaves as expected with the plot title too.

R/ggplotly.R
doorisajar: Interestingly, this variant fails the test, even though it definitely returns `FALSE` when passed a `NULL` or `logical(0)` explicitly. Not sure what's different about how `axisTitleText` is being received. For now I'll get it passing the test.

doorisajar: Totally forgot about `rlang::is_empty` for dealing with this stuff. That seems to do the trick.

cpsievert: plotly currently doesn't have an rlang dependency, so I'd rather use `length(x) == 0` than `rlang::is_empty()`

cpsievert:
What happened in 3d99fad? Can you fix the diff on R/ggplotly.R?

doorisajar:
> 
> 
> What happened in [3d99fad](https://github.com/ropensci/plotly/commit/3d99fad9876203d48a113f76b1638af59119e95b)? Can you fix the diff on R/ggplotly.R?

Looks like there's no line endings specification in the Rproj file, so when I switched from editing things in the GitHub interface to editing them in RStudio on Windows, the line endings got changed and I didn't notice. Just attempted to revert that in de198b4...

R/ggplotly.R
doorisajar: `rlang` was already in the imports for `eval_tidy`: https://github.com/ropensci/plotly/blob/master/DESCRIPTION#L43

cpsievert:
I'd still prefer to use `length()` directly over `rlang::is_empty()` as I more regularly use the former

doorisajar:
> I'd still prefer to use `length()` directly over `rlang::is_empty()` as I more regularly use the former

In case you didn't see the commit, I did make this change.

cpsievert:
Thanks for contributing this fix! It had some overlap with #1773 so we incorporated it there

PR #1771, opened 2020-05-18
Upgrade to plotly.js v1.54.1

cpsievert:


nicolaskruchten:
Awesome! I'm glad we finally got this "Plotly team submits the PR" process run through once :)

You can expect a follow-up PR for 1.54.2 in the coming days, and then 1.55.0 will come out in a few weeks. No new trace types are expected for a while, so hopefully this process remains smooth.

@cpsievert what's your approach towards CRAN and release announcements? What we like to do with Plotly.py is that every time we write up the docs at plotly.com/python for the new features, we do a Community Forum post like [these ones](https://community.plotly.com/tag/releases). We recently "caught up" on the R side with our docs and belatedly put out [this post](https://community.plotly.com/t/announcing-plotly-r-4-9-2/37775) as a "roundup". So now that there are new features in Plotly.R since that post, @jdamiba will port the Python docs (for unified hover, range breaks and shape drawing, basically) over to plotly.com/r and do another forum post to announce the new features (probably after 1.54.2 is merged). If that could be synced with a CRAN update that's cool but I don't see it as critical, just curious about your thoughts :)

cpsievert:
It's best to be somewhat conservative about CRAN releases -- I usually aim for about a 2 month release cycle (unless critical patch fixes surface, of course)

nicolaskruchten:
OK cool. Nothing critical is in the works ATM :) Would appreciate a retweet on the community forum post when it comes out!

PR #1773, opened 2020-05-19
ggplotly() failed when labels or title were a factor, because nchar()…

### LINKED ISSUE #1772: Conversion of discrete labels fails if they come from a factor
This StackOverflow question https://stackoverflow.com/q/61888967/2554330 motivated this report.

If the labels in `scale_x_discrete` are a factor, `ggplot` works fine.  This is okay:

    ggplot(diamonds, aes(cut)) + 
      geom_bar() + 
      scale_x_discrete("Cut", labels=factor(letters[1:5]))

However, `ggplotly()` fails:

``` r
ggplotly(
  ggplot(diamonds, aes(cut)) + 
  geom_bar() + 
  scale_x_discrete("Cut", labels=factor(letters[1:5]))
)
#> Error in nchar(axisObj$ticktext): 'nchar()' requires a character vector
```

Apparently an `as.character()` call is needed when `ticktext` is constructed.

dmurdoch:
… doesn't like factors.  This PR adds `nchar0()`, which converts its argument to character before counting chars, and fixes #1772.

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -82,10 +82,6 @@ ggplotly.plotly <- function(p = ggplot2::last_plot(), width = NULL, height = NUL
   p
 }
 
-# nchar() needs a character vector; sometimes x will be a
-# factor
-nchar0 <- function(x, ...) nchar(as.character(x), ...)
-
 #' @export
 ggplotly.ggmatrix <- function(p = ggplot2::last_plot(), width = NULL,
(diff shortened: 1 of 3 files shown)


cpsievert:
Thanks! It'd be great to account for factors as well as zero length input (see #1731). Would you mind modifying `nchar0()` to account for that case as well (and give it a more suitable name)?

PR #1779, opened 2020-05-29
Geom errorbar issure 1751

### LINKED ISSUE #1751: geom_errorbar is not rendered unless the y aes is set
In ggplot2, the only aesthetics needed for geom_errorbar are ymin and ymax. However, ggplotly fails unless I also specify the y aesthetic.

``` r
library(ggplot2)
library(plotly)

data <- data.frame(auc=c(0.268707482993197,0.571428571428571), 
                   sup=c(0.407680628614317,0.648343533190079), 
                   inf=c(0.129734337372078,0.494513609667063), 
                   Names = c("Firmicutes","Spirochaetes"))


ggplot(data, aes(Names)) +
  geom_errorbar(aes(ymin = inf, ymax = sup))    
```

``` r
plotly::ggplotly()
```
image

``` r

ggplot(data, aes(Names)) +
  geom_errorbar(aes(y = auc, ymin = inf, ymax = sup))    
```

``` r
plotly::ggplotly()
```

image

<sup>Created on 2020-04-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

trekonom:
This PR fixes #1751 . As far as I get it the bug arises in `make_error`. If the `y` aes is not set, `data[["y"]]` is `NULL`. In that case empty vectors are assigned to `arrayminus `and `array`. To prevent this I added a check if `data[["y"]]` is NULL and set `data[["y"]]` equal to (`data[["ymin "]] + data[["ymax"]]) / 2` if `TRUE`. I also added some tests.


trekonom:
I updated my PR to add a solution for the issue that rendering `geom_errorbar ` breaks with flipped aesthetics. I attached a reprex which shows the problem:

``` r
library(ggplot2)
library(plotly)

# Errobar breaks with flipped aesthetics
gp <- iris %>% 
  dplyr::group_by(Species) %>% 
  dplyr::summarise_if(is.numeric, list(m = mean, q1 = ~ quantile(.x, .25), q3 = ~ quantile(.x, .75))) %>% 
  ggplot(. , aes(y = Species, xmin = Sepal.Width_q1, xmax = Sepal.Width_q3)) +
  geom_errorbar()
gp
```



``` r
ggplotly(gp)
#> Error in `$<-.data.frame`(`*tmp*`, "width", value = numeric(0)): replacement has 0 rows, data has 3
```

<sup>Created on 2020-05-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/layers2traces.R
```diff
@@ -873,7 +877,11 @@ geom2trace.GeomTile <- function(data, params, p) {
 
 #' @export
 geom2trace.GeomErrorbar <- function(data, params, p) {
-  make_error(data, params, "y")
+  if (params[["flipped_aes"]]) {
```
cpsievert, line 880: Is this a new ggplot2 param? I haven't noticed it before (where does it come from?) Could you provide an example of where it's relevant? Preferably, provide a very small reprex that we can include as a comment to explain why this is here

R/layers2traces.R
```diff
@@ -951,6 +959,10 @@ hover_on <- function(data) {
 
 # make trace with errorbars
 make_error <- function(data, params, xy = "x") {
+  # if xy is NULL: set xy to mean of xy_min and xy_max
+  if (is.null(data[[xy]])) {
+    data[[xy]] <- (data[[paste0(xy, "min")]] + data[[paste0(xy, "max")]]) / 2  
+  }
```
cpsievert, line 965:
```suggestion
  data[[xy]] <- data[[xy]] %||% (data[[paste0(xy, "min")]] + data[[paste0(xy, "max")]]) / 2  
```

trekonom: We need another pair of parentheses on the RHS:
```suggestion
data[[xy]] <- data[[xy]] %||% ((data[[paste0(xy, "min")]] + data[[paste0(xy, "max")]]) / 2)
```

R/layers2traces.R
trekonom: Hi Carson. 

You are right. This param was introduced in ggplot2 3.3.0 to support [bi-directional geoms](https://www.tidyverse.org/blog/2020/03/ggplot2-3-3-0/#bi-directional-geoms-and-stats). This functionality is implemented in [has_flipped_aes, ...](https://ggplot2.tidyverse.org/reference/bidirection.html).

I haven't realized that this is a more general change. In so far my PR offers only a fix for the case of `GeomErrorbar `but no general solution. I added a comment to explain why this is here together with a small reprex.

```suggestion
# Support of bi-directional GeomErrorbar introduced with ggplot2 3.3.0:
# g <- ggplot() + geom_errorbar(aes(y = "A", xmin = 1, xmax = 2))
# ggplotly(g)
if (params[["flipped_aes"]]) { 
```

trekonom:
After getting visual testing inside RStudio and via docker to work I added visual tests to the PR.

cpsievert:
Thank you @trekonom, sorry for the wait!

PR #1782, opened 2020-06-01
Add support for ggalt::geom_dumbbell.

### LINKED ISSUE #1454: Add support for geom_dumbbell
Please briefly describe your problem and what output you expect. If you have a question, please don't use this form, but instead ask on the community forum <http://community.plot.ly/c/api/r> or stackoverflow <http://stackoverflow.com>.

Please include a minimal reprex. The goal of a reprex is to make it as easy as possible for me to recreate your problem so that I can fix it. If you've never heard of a reprex before, start by reading <https://github.com/tidyverse/reprex#readme>, and follow the advice further down the page. Do NOT include session info unless it's explicitly asked for, or you've used `reprex::reprex(..., si = TRUE)` to hide it away.  

Delete these instructions once you have read them.

---

Brief description of the problem

```r
# insert reprex here
```

trekonom:
This PR adds support for `ggalt::geom_dumbbell`  and closes #1454 . 

- The implementation follows `ggalt::GeomDumbbell`.

- Tooltips only show up if hovering over the points and differ for the left and right point. To this end I added a util function `make_hovertext `which overwrites the hovertext set by `layers2traces`.

- I also added some tests and tried to add a visual test (which however failed when running the test while vdiffr::manage_cases showed me a successful doppelganger). Therefore the visual test is commented out in the test.


R/layers2traces.R
```diff
@@ -608,6 +608,65 @@ to_basic.GeomQuantile <- function(data, prestats_data, layout, params, p, ...){
   dat
 }
 
+#' @export
+to_basic.GeomDumbbell <- function(data, prestats_data, layout, params, p, ...) {
+  # Adds support for ggalt::geom_dumbbell
+  # Implementation follows ggalt::GeomDumbbell
+  # Setup data for left points
+  points.x <- data
```
cpsievert, line 616: Please use `_` instead of `.` for separating names

R/layers2traces.R
```diff
@@ -608,6 +608,65 @@ to_basic.GeomQuantile <- function(data, prestats_data, layout, params, p, ...){
...
+      prefix_class(points.xend, "GeomPoint")
+    )
+  } else {
+    list(
+      prefix_class(dot_df, "GeomPath"),
+      prefix_class(data, "GeomPath"),
+      prefix_class(points.x, "GeomPoint"),
+      prefix_class(points.xend, "GeomPoint")
+    )
+  }
```
cpsievert, line 667:
```suggestion
  list(
    if (is.null(params$dot_guide) || !params$dot_guide) prefix_class(dot_df, "GeomPath"),
    prefix_class(data, "GeomPath"),
    prefix_class(points.x, "GeomPoint"),
    prefix_class(points.xend, "GeomPoint")
  )
```

R/layers2traces.R
```diff
@@ -1125,3 +1184,41 @@ lty2dash <- function(x) {
...
+      txt <- x[[paste0(aesName, "_plotlyDomain")]] %||% x[[aesName]]
+      suffix <- tryNULL(format(txt, justify = "none")) %||% ""
+      if (inherits(x, "GeomBar") && identical(aesName, "y")) {
+        suffix <- format(x[["ymax"]] - x[["ymin"]], justify = "none")
+      }
+      x$hovertext <- paste0(x$hovertext, prefix, suffix)
+    }
+    x$hovertext <- x$hovertext %||% ""
+    x
+  }, data, hoverTextAes)
```
cpsievert, line 1222: Instead of copy/pasting this from the body of `layers2traces()` let's extract this bit into it's own function that we can use it both places

trekonom:
Hi Carson. First. Sorry for messing up the commit history. 

- I folllowed your suggestions for simplifying the code, removing the duplicated code and renaming the variables. 
- After commiting I realized that I thereby introduced a bug. The latest commits fix this bug. I also added an addtional test to check that.
- Finally I removed some code duplication by adding a function `segement2path `. 

Best, Stefan

trekonom:
I just added the visual tests to the PR. However, running the visual tests against the docker image failed with the message "there is no package called 'ggalt'". If I should guess ... this probably happens because `ggalt` is not included in the docker image.

cpsievert:
Ah, yea, let's add `ggalt` to `Suggests` in the `DESCRIPTION` file

PR #1787, opened 2020-06-07
Add modeBarButtonsToKeep argument

### LINKED ISSUE #1740: New feature request: a modeBarButtonsToKeep argument in the config function
Currently plotly has a function to remove listed mode bar buttons. 

It'd be nice if it had a function to keep listed mode bar buttons, and to drop the rest. 

Otherwise the code is quite long-winded if a user only wants to keep one button, such as the camera. 

This argument could be called `modeBarButtonsToKeep`

```r
library(dplyr)
library(ggplot2)

plot <- ggplot(iris) +
  geom_point(aes(Sepal.Width, Sepal.Length))

plotly::ggplotly(plot) %>% 
  plotly::config(modeBarButtonsToRemove = list("zoom2d", "pan2d", "zoomIn2d", "zoomOut2d", "autoScale2d","resetScale2d", "hoverClosestCartesian", "hoverCompareCartesian", "sendDataToCloud", "toggleHover", "resetViews", "toggleSpikelines", "resetViewMapbox", "toggleSpikelines", "resetViewMapbox", "lasso2d", "select2d"))

```

trekonom:
This PR adds a `modeBarButtonsToKeep `argument to the configuration options and closes #1740 .


R/layout.R
```diff
@@ -162,9 +162,39 @@ config <- function(p, ..., cloud = FALSE, showSendToCloud = cloud, locale = NULL
   
   args <- list(...)
   if ("collaborate" %in% names(args)) warning("The collaborate button is no longer supported")
+  if ("modeBarButtonsToKeep" %in% names(args)) args <- modebar_keep(args)
```
cpsievert, line 165: I'd prefer to make this an explicit argument in `config()` (after `...`, since it's not an official plotly.js config attribute)

R/layout.R
```diff
@@ -162,9 +162,39 @@ config <- function(p, ..., cloud = FALSE, showSendToCloud = cloud, locale = NULL
...
+  modebar_names <- modebar_names()
+  x <- setdiff(modebar_names, unlist(args[["modeBarButtonsToKeep"]]))
+  args[["modeBarButtonsToRemove"]] <- as.list(x)
+  # Remove modeBarButtonsToKeep from args to prevent warning
+  args[["modeBarButtonsToKeep"]] <- NULL
+  args
+}
+
+# List of mode bar names from
+# https://github.com/plotly/plotly.js/blob/master/src/components/modebar/buttons.js
```
cpsievert, line 186: It'd be cool if we could modify the [build script](https://github.com/ropensci/plotly/blob/master/inst/plotlyjs.R) I use when updating plotly.js to programmatically update this object. I imagine you might be use something like the V8 package to source this file and get the button names as an R object?

R/layout.R
trekonom: Okay. I see. I just added a [commit](https://github.com/ropensci/plotly/pull/1787/commits/ea15947839b146b850619fd3eb32f195db30a9f2) to fix this.

R/layout.R
trekonom: Yep. Sure. To get the list of modebar names I simply did some kind of text-mining or scraping like so:

```
get_modebar_names <- function(url = "https://raw.githubusercontent.com/plotly/plotly.js/master/src/components/modebar/buttons.js") {
  modebar_names <- readLines(url)
  modebar_names <- stringr::str_trim(modebar_names)
  modebar_names <- modebar_names[stringr::str_detect(modebar_names, "^modeBarButtons\\.")]
  stringr::str_match(modebar_names, "modeBarButtons\\.(\\w+)")[, 2]
}
modebar_names <- get_modebar_names()
```
If this is fine for you I can add it to the build script. In that case I would rewrite it using base R functions to get rid of the `stringr `dependency.

R/layout.R
cpsievert: Great, thanks!

R/layout.R
cpsievert: I would prefer to fo the V8 route if it's possible, but if it's not possible or too much trouble, we can go this route. Also, we should make it part of the the build script (as I linked to above), and write some sensible unit tests to check the value of `modebar_names` (i.e., it's a character vector that contains at least the default modebar buttons like `zoom2d`, etc)

Issue #1577, opened 2019-07-13
x axis problem when using facets with a unique value on the x axis


paobranco:
Hi, 
Great package!

I found an issue when using facet_wrap and facet_grid.
The problem is related with the existence of a unique value in the x axis. When this happens, only the first panel shows the x axis correctly.
Here is a reproducible example:

```r
data(iris)
ds <- iris[1:10,]
ds$var <- as.factor(c(rep("A", 5), rep("B",5)))
p <- ggplot(ds, aes(x=Species, y=Petal.Length))+facet_grid(~var)+geom_point()
p
```
This produces this plot:
![Rplot1](https://user-images.githubusercontent.com/3248426/61164115-7902e680-a50a-11e9-803c-cec9feffc499.png)

However, when using ggplotly
```r
ggplotly(p)
```
This happens:
![Rplot2](https://user-images.githubusercontent.com/3248426/61164132-92a42e00-a50a-11e9-9dcd-b5714ae92752.png)

The x axis of the first panel is OK, but all the other panels will show numbers on the x axis.

Thank you!

cpsievert:
Fixed via #1788

PR #1788, opened 2020-06-07
Fix axis facets

### LINKED ISSUE #1720: ggplotly breaks discrete x-axis with facet_yyyy in panels > 1 with only one category
ggplotly breaks a discrete axis with` facet_yyyy` when there is only one category present in a  panel, except for the case that this happens to be the first panel. I stumbled over this bug based on a question on [Stackoverflow](https://stackoverflow.com/questions/60606101/x-axis-error-with-ggplotly-and-facet-wrap#comment107224862_60606101):
``` r
library(plotly)

d <- data.frame(cat = c("A", "A", "A"), pan = paste("Panel", c(1, 2, 2)))
ggplot(d, aes(cat)) +
  geom_bar() +
  facet_wrap(~pan)
```

```
ggplotly()
```

facet-breaks-axis

<sup>Created on 2020-06-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

trekonom:
This PR fixes a bug in `verify_attr` which closes #1720 and #1577 .

Adding the fix the reprex from #1720 is rendered like so

grafik

and the reprex from #1577 like so:

grafik

### DIFF
--- a/tests/testthat/test-facet-axis.R
+++ b/tests/testthat/test-facet-axis.R
@@ -1,0 +1,11 @@
+test_that("ggplotly does not break discrete x-axis with facet_yyyy in panels > 1 with only one category", {
+  d <- data.frame(cat = c("A", "A", "A"), pan = paste("Panel", c(1, 2, 2)))
+  gp <- ggplot(d, aes(cat)) +
+    geom_bar() +
+    facet_wrap(~pan)
+  L <- plotly_build(ggplotly(gp))
+  # tickvals, ticktext and categoryarray have class 'AsIs'
+  expect_equal(class(L$x$layout$xaxis2$tickvals), "AsIs")
+  expect_equal(class(L$x$layout$xaxis2$ticktext), "AsIs")
+  expect_equal(class(L$x$layout$xaxis2$categoryarray), "AsIs")
+})
--- a/tests/testthat/test-facet-axis.R
+++ b/tests/testthat/test-facet-axis.R
@@ -5,7 +5,11 @@ test_that("ggplotly does not break discrete x-axis with facet_yyyy in panels > 1
     facet_wrap(~pan)
   L <- plotly_build(ggplotly(gp))
   # tickvals, ticktext and categoryarray have class 'AsIs'
-  expect_equal(class(L$x$layout$xaxis2$tickvals), "AsIs")
-  expect_equal(class(L$x$layout$xaxis2$ticktext), "AsIs")
-  expect_equal(class(L$x$layout$xaxis2$categoryarray), "AsIs")
+  lapply(L$x$layout[c("xaxis", "xaxis2")], function(axis) {
+    expect_s3_class(axis$tickvals, "AsIs")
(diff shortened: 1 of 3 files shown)


tests/testthat/test-facet-axis.R
```diff
@@ -0,0 +1,11 @@
+test_that("ggplotly does not break discrete x-axis with facet_yyyy in panels > 1 with only one category", {
+  d <- data.frame(cat = c("A", "A", "A"), pan = paste("Panel", c(1, 2, 2)))
+  gp <- ggplot(d, aes(cat)) +
+    geom_bar() +
+    facet_wrap(~pan)
+  L <- plotly_build(ggplotly(gp))
+  # tickvals, ticktext and categoryarray have class 'AsIs'
+  expect_equal(class(L$x$layout$xaxis2$tickvals), "AsIs")
+  expect_equal(class(L$x$layout$xaxis2$ticktext), "AsIs")
+  expect_equal(class(L$x$layout$xaxis2$categoryarray), "AsIs")
```
cpsievert, line 10:
```suggestion
  lapply(L$x$layout[c("xaxis", "xaxis2")], function(axis) {
    expect_s3_class(axis$tickvals, "AsIs")
    expect_s3_class(axis$ticktext, "AsIs")
    expect_true(axis$ticktext == "A")
    expect_s3_class(axis$categoryarray, "AsIs")
    expect_true(axis$categoryarray == "A")
  })
```

cpsievert:
Looks great, thank you! If you agree with the suggested change, please commit it and add a news item, then I'll merge

BTW, have you tried to create and/or [run visual tests](https://github.com/ropensci/plotly/blob/master/CONTRIBUTING.md#running-visual-tests-via-docker)? I think we can get away with not having one here, but it'd be great to have them in some of your other PRs

trekonom:
Just commited your changes and added a news item.

Regarding the visual tests:

- Visual tests and svgs are already present for PRs #1779 and #1782 .  
- However the visual tests always failed while vdiffr::manage_cases showed successful doppelganger.
- Therefore I decided to comment them out in my commits.

I also tried running the visual tests via docker ... 

- Following your instructions , i.e. running `docker run -v $(pwd):/home/plotly --privileged -p 3838:3838 cpsievert/plotly-orca` always resulted in the error `invalid reference format`.
- But as I used docker for the first time maybe I simply got something wrong.

cpsievert:
Hmm, are you in windows powershell? https://stackoverflow.com/a/51379726/1583084

trekonom:
(: Thanks a lot. It was the Powershell issue. Now the docker works like a charm. BTW: I also figured out my issue with visual testing inside RStudio. After running the tests via `devtools::test() `(as you suggested in the contributing guidelines) instead of pressing the "Run Tests" button my visual tests passed.

cpsievert:
Ahh, yea, sorry, I should've mentioned before that the visual tests should only really only be run via docker

PR #1791, opened 2020-06-12
Add fix to take account of linebreaks in yaxis ticklabels.

trekonom:
This PR proposes a fix to take account of linebreaks in yaxis ticklabels.

At present `ggplotly` does not take linebreaks in y-axis labels into account when computing the margins. Hence, the left margin is set too wide in that case. 

The issue can be illustrated with this reprex:

```
library(plotly)

d <- data.frame(x = c(1, 2), y = c("dfdfjdfdf\ndjdfkjdfdklj\nsldflkfdlf", "dfdfjdfdf"))

ggplot(d, aes(x, y)) +
  geom_bar(stat = "identity")

ggplotly()
```

At present this is rendered with a lot of white space to the left of the axis labels:
 
grafik

With the proposed fix the reprex is rendered like so:

grafik

As a workaround to this issue one can force `automargin` to do its job by setting the left margin to a small number in the layout options like so:

```
ggplotly() %>%
    layout(margin = list(l = 0))
```


R/ggplotly.R
```diff
@@ -806,7 +806,12 @@ gg2list <- function(p, width = NULL, height = NULL,
       
       # do some stuff that should be done once for the entire plot
       if (i == 1) {
-        axisTickText <- longest_element(axisObj$ticktext)
+        # Split ticktext elements by "\n"  to account for linebreaks
+        axisTickText <- sapply(as.character(axisObj$ticktext), strsplit, split = "\n", USE.NAMES = FALSE)
```
cpsievert, line 810: Is the `sapply()` necessary here? Couldn't it just be `strsplit(axisObj$ticktext, split = "\n", USE.NAMES = FALSE)`

trekonom: Yep. You are right. `sapply` isn't necessary. I will fix that and also try to add some tests.

PS: I will keep your hint on the use `sapply` in mind.

trekonom: Removed the unnecessary `sapply `and added some tests. Visual tests have been run against the orca image.

tests/testthat/test-ticktext-linebreaks.R
```diff
@@ -0,0 +1,64 @@
...
+  
+  gglayout$margin[["l"]] + axisObj$ticklen +
+    bbox(axisTickText, 0, axisObj$tickfont$size)[["width"]] +
+    bbox("y", 90, unitConvert(axisTitle, "pixels", "width"))[["width"]]
+}
+
+expect_margin <- function(L, gg, ticktext) {
+  margin_l <- comp_margin(gg, ticktext)
+  expect_equal(round(L$layout$margin$l, 10), round(margin_l, 10))
+}
```
cpsievert, line 31: I appreciate you going down this road, but just FYI, I think the visual tests would have been sufficient here.

R/ggplotly.R
```diff
@@ -806,7 +806,12 @@ gg2list <- function(p, width = NULL, height = NULL,
       
       # do some stuff that should be done once for the entire plot
       if (i == 1) {
-        axisTickText <- longest_element(axisObj$ticktext)
+        # Split ticktext elements by "\n"  to account for linebreaks
+        axisTickText <- strsplit(as.character(axisObj$ticktext), split = "\n")
```
cpsievert, line 810:
```suggestion
        axisTickText <- strsplit(as.character(axisObj$ticktext), split = "\n", fixed = TRUE)
```

R/ggplotly.R
```diff
@@ -806,7 +806,12 @@ gg2list <- function(p, width = NULL, height = NULL,
...
+        # Get longest element for each tick
+        axisTickText <- lapply(axisTickText, longest_element)
+        # Longest element of all ticks
+        axisTickText <- longest_element(axisTickText)
```
cpsievert, line 814: I think this would be slightly faster

```suggestion
        axisTickText <- longest_element(unlist(axisTickText))
```

tests/testthat/test-ticktext-linebreaks.R
trekonom: As I just started with the contributing and testing business this was a good exercise. (;

PR #1800, opened 2020-06-23
Update dplyr functions to use non-standard evaluation

### LINKED ISSUE #1783: Warning: `arrange_()` is deprecated as of dplyr 0.7.0. Please use `arrange()` instead.
I update dplyr to 1.0.0 and the following message appears:

`Warning: `arrange_()` is deprecated as of dplyr 0.7.0. Please use `arrange()` instead. See vignette('programming') for more help`

julianstanley:
Replaced all instances of depreciated standard evaluation dplyr verbs with their NSE equivalents.

Fixes associated depreciation warnings, so closes #1783. For example,

Before:

```R
> p <- ggplot(txhousing, aes(x = date, y = median, group = city)) +
+   geom_line(alpha = 0.3)
> ggplotly(p) %>% filter(city == "Houston")
Warning message:
`filter_()` is deprecated as of dplyr 0.7.0.
Please use `filter()` instead.
See vignette('programming') for more help
```

After:

```R
> p <- ggplot(txhousing, aes(x = date, y = median, group = city)) +
+   geom_line(alpha = 0.3)
> ggplotly(p) %>% filter(city == "Houston")
```

This PR introduces a util function called `as_quosures_explicit`. If/when [issue 1009 in r-lang](https://github.com/r-lib/rlang/issues/1009) is resolved, just replace all instances of `as_quosures_explicit` with `rlang::as_quosures`.

### DIFF
--- a/R/add.R
+++ b/R/add.R
@@ -334,7 +334,7 @@ add_sf <- function(p, ..., x = ~x, y = ~y, data = NULL, inherit = TRUE) {
       y = y,
       `_bbox` = bbox,
       set = set,
-      data = if ("group" %in% names(d[[i]])) group_by_(d[[i]], "group", add = TRUE) else d[[i]], 
+      data = if ("group" %in% names(d[[i]])) group_by(d[[i]], rlang::sym("group"), .add = TRUE) else d[[i]], 
       inherit = inherit
     )
     args <- modify_list(args, attrs)
--- a/R/add.R
+++ b/R/add.R
@@ -334,7 +334,7 @@ add_sf <- function(p, ..., x = ~x, y = ~y, data = NULL, inherit = TRUE) {
       y = y,
       `_bbox` = bbox,
       set = set,
-      data = if ("group" %in% names(d[[i]])) group_by_add(d[[i]], !!rlang::sym("group"), add = TRUE) else d[[i]], 
+      data = if ("group" %in% names(d[[i]])) group_by_add(d[[i]], !!rlang::sym("group")) else d[[i]], 
       inherit = inherit
     )
     args <- modify_list(args, attrs)
--- a/R/add.R
+++ b/R/add.R
@@ -334,7 +334,7 @@ add_sf <- function(p, ..., x = ~x, y = ~y, data = NULL, inherit = TRUE) {
       y = y,
       `_bbox` = bbox,
       set = set,
-      data = if ("group" %in% names(d[[i]])) group_by(d[[i]], !!rlang::sym("group"), .add = TRUE) else d[[i]], 
+      data = if ("group" %in% names(d[[i]])) group_by_add(d[[i]], !!rlang::sym("group"), add = TRUE) else d[[i]], 
       inherit = inherit
     )
     args <- modify_list(args, attrs)
--- a/R/add.R
+++ b/R/add.R
@@ -334,7 +334,7 @@ add_sf <- function(p, ..., x = ~x, y = ~y, data = NULL, inherit = TRUE) {
       y = y,
       `_bbox` = bbox,
       set = set,
-      data = if ("group" %in% names(d[[i]])) group_by(d[[i]], rlang::sym("group"), .add = TRUE) else d[[i]], 
+      data = if ("group" %in% names(d[[i]])) group_by(d[[i]], !!rlang::sym("group"), .add = TRUE) else d[[i]], 
       inherit = inherit
     )
     args <- modify_list(args, attrs)
--- a/R/imports.R
+++ b/R/imports.R
@@ -24,11 +24,23 @@ NULL
 #' @export
 dplyr::mutate
 
+#' @importFrom dplyr mutate_
+#' @name mutate_
+#' @rdname reexports
+#' @export
+dplyr::mutate_
+
 #' @importFrom dplyr transmute
 #' @name transmute
 #' @rdname reexports
 #' @export
 dplyr::transmute
+
+#' @importFrom dplyr transmute_
+#' @name transmute_
+#' @rdname reexports
+#' @export
+dplyr::transmute_
 
 #' @importFrom dplyr select
 #' @name select
@@ -36,17 +48,35 @@ dplyr::transmute
 #' @export
 dplyr::select
 
+#' @importFrom dplyr select_
+#' @name select_
+#' @rdname reexports
+#' @export
+dplyr::select_
+
 #' @importFrom dplyr rename
 #' @name rename
 #' @rdname reexports
 #' @export
 dplyr::rename
 
+#' @importFrom dplyr rename_
+#' @name rename_
+#' @rdname reexports
+#' @export
+dplyr::rename_
+
 #' @importFrom dplyr group_by 
 #' @name group_by
 #' @rdname reexports
 #' @export
 dplyr::group_by
+
+#' @importFrom dplyr group_by_
+#' @name group_by_
+#' @rdname reexports
+#' @export
+dplyr::group_by_
 
 #' @importFrom dplyr groups
 #' @name groups
@@ -66,22 +96,47 @@ dplyr::ungroup
 #' @export
 dplyr::summarise
 
+#' @importFrom dplyr summarise_
+#' @name summarise_
+#' @rdname reexports
+#' @export
+dplyr::summarise_
+
 #' @importFrom dplyr do
 #' @name do
 #' @rdname reexports
 #' @export
 dplyr::do
 
+#' @importFrom dplyr do_
+#' @name do_
+#' @rdname reexports
+#' @export
+dplyr::do_
+
 #' @importFrom dplyr arrange
 #' @name arrange
 #' @rdname reexports
 #' @export
 dplyr::arrange
+
+#' @importFrom dplyr arrange_
+#' @name arrange_
+#' @rdname reexports
+#' @export
+dplyr::arrange_
+
 #' @importFrom dplyr distinct
 #' @name distinct
 #' @rdname reexports
 #' @export
 dplyr::distinct
+
+#' @importFrom dplyr distinct_
+#' @name distinct_
+#' @rdname reexports
+#' @export
+dplyr::distinct_
 
 #' @importFrom dplyr slice
 #' @name slice
@@ -89,11 +144,23 @@ dplyr::distinct
 #' @export
 dplyr::slice
 
+#' @importFrom dplyr slice_
+#' @name slice_
+#' @rdname reexports
+#' @export
+dplyr::slice_
+
 #' @importFrom dplyr filter
 #' @name filter
 #' @rdname reexports
 #' @export
 dplyr::filter
+
+#' @importFrom dplyr filter_
+#' @name filter_
+#' @rdname reexports
+#' @export
+dplyr::filter_
 
 # waiting on https://github.com/tidyverse/tidyr/pull/229
 #
--- a/R/imports.R
+++ b/R/imports.R
@@ -24,23 +24,11 @@ NULL
 #' @export
 dplyr::mutate
 
-#' @importFrom dplyr mutate_
-#' @name mutate_
-#' @rdname reexports
-#' @export
-dplyr::mutate_
-
 #' @importFrom dplyr transmute
 #' @name transmute
 #' @rdname reexports
 #' @export
 dplyr::transmute
-
-#' @importFrom dplyr transmute_
-#' @name transmute_
-#' @rdname reexports
-#' @export
-dplyr::transmute_
 
 #' @importFrom dplyr select
 #' @name select
@@ -48,35 +36,17 @@ dplyr::transmute_
 #' @export
 dplyr::select
 
-#' @importFrom dplyr select_
-#' @name select_
-#' @rdname reexports
-#' @export
-dplyr::select_
-
 #' @importFrom dplyr rename
 #' @name rename
 #' @rdname reexports
 #' @export
 dplyr::rename
 
-#' @importFrom dplyr rename_
-#' @name rename_
-#' @rdname reexports
-#' @export
-dplyr::rename_
-
 #' @importFrom dplyr group_by 
 #' @name group_by
 #' @rdname reexports
 #' @export
 dplyr::group_by
-
-#' @importFrom dplyr group_by_
-#' @name group_by_
-#' @rdname reexports
-#' @export
-dplyr::group_by_
 
 #' @importFrom dplyr groups
 #' @name groups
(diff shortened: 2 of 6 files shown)


julianstanley:
Build 4813.1 and 4813.2 passed (devel and oldrel). 4813.3 failed:

```r
══ Failed ══════════════════════════════════════════════════════════════════���═══
── 1. Error: Complex example works (@test-plotly.R#179)  ───────────────────────
Column `.add` can't be modified because it's a grouping variable
Backtrace:
  1. plotly::plot_ly(., x = ~date, y = ~median)
  1. dplyr::group_by(., city)
  1. plotly::add_lines(., alpha = 0.2, name = "Texan Cities", hoverinfo = "none")
  8. dplyr::group_by(., date)
 11. plotly::group_by_.plotly(...)
 13. dplyr:::group_by.data.frame(d, !!!additional_args, .add = add)
 14. dplyr::group_by_prepare(.data, ..., add = add)
 15. dplyr:::add_computed_columns(.data, new_groups)
 17. dplyr:::mutate.tbl_df(.data, !!!mutate_vars)
 18. dplyr:::mutate_impl(.data, dots, caller_env())
```

Related to [this StackOverflow post](https://stackoverflow.com/questions/50029128/using-mutate-returns-error-about-cant-modify-because-its-a-grouping-variable), maybe?

R/utils.R
```diff
@@ -139,7 +139,7 @@ crosstalk_key <- function() ".crossTalkKey"
 # arrange data if the vars exist, don't throw error if they don't
 arrange_safe <- function(data, vars) {
   vars <- vars[vars %in% names(data)]
-  if (length(vars)) dplyr::arrange_(data, .dots = vars) else data
+  if (length(vars)) dplyr::arrange(data, !!!lapply(vars, rlang::sym)) else data
```
cpsievert, line 142:
```suggestion
  if (length(vars)) dplyr::arrange(data, !!!rlang::syms(vars)) else data
```

R/utils.R
```diff
@@ -1154,3 +1154,11 @@ longest_element <- function(x) {
...
     ""
 }
+
+# Apply rlang::as_quosure across a list or vector, but explicitly pass env
+# and expr to as_quosure, due to a bug in as_quosures when dealing with 
+# lazy objects
+as_quosures_explicit <- function(vec) {
+  lapply(vec, 
+         FUN = function(x) rlang::as_quosure(x$expr, x$env))
+}
```
cpsievert, line 1164: I think I'd want to make it more explicit that this function is for coercing dots to quosures


```r
dots_as_quosures <- function(x) {
  if (!inherits(x, "lazy_dots")) {
    stop("Expected lazy dots")
  }
  lapply(vec, function(x) rlang::as_quosure(x$expr, x$env))
}
```

R/plotly_build.R
```diff
@@ -561,7 +561,7 @@ train_data <- function(data, trace) {
     idx2 <- seq.int(2, NROW(dat), by = 2)
     dat[idx2, "x"] <- data[["xend"]]
     dat[idx2, "y"] <- data[["yend"]]
-    data <- dplyr::group_by_(dat, ".plotlyGroupIndex", add = TRUE)
+    data <- dplyr::group_by(dat, !!rlang::sym(".plotlyGroupIndex"), .add = TRUE)
```
cpsievert, line 564: I don't think this would work for dplyr < 1.0, right? It might be useful to have a helper like:

```r
group_by_add <- function(..., add = TRUE) {
  if (packageVersion('dplyr') >= '1.0') {
      dplyr::group_by(...,  .add = add)
  } else {
      dplyr::group_by(...,  add = add)
  } 
}
```

R/plotly_data.R
```diff
@@ -121,10 +121,11 @@ ungroup.plotly <- function(x, ...) {
 #' @export
 group_by_.plotly <- function(.data, ..., .dots, add = FALSE) {
   d <- plotly_data(.data)
-  d2 <- dplyr::group_by_(d, .dots = lazyeval::all_dots(.dots, ...), add = add)
+  additional_args <- as_quosures_explicit(lazyeval::all_dots(.dots, ...))
+  d2 <- dplyr::group_by(d, !!!additional_args, .add = add)
   # add crosstalk key as a group (to enable examples like demos/highlight-pipeline.R)
   if (crosstalk_key() %in% names(d)) {
-    d2 <- dplyr::group_by_(d2, crosstalk_key(), add = TRUE)
+    d2 <- dplyr::group_by(d2, !!crosstalk_key(), .add = TRUE)
```
cpsievert, line 128:
```suggestion
    d2 <- dplyr::group_by(d2, !!rlang::sym(crosstalk_key()), .add = TRUE)
```

R/imports.R
```diff
@@ -24,60 +24,30 @@ NULL
 #' @export
 dplyr::mutate
 
-#' @importFrom dplyr mutate_
-#' @name mutate_
-#' @rdname reexports
-#' @export
-dplyr::mutate_
```
cpsievert, line 31: What's your reasoning for removing these exports?

julianstanley: I just thought that importing the SE verbs would be unnecessary since we were switching to NSE-only. Should I put them back in?

cpsievert: These are "re-exports" meaning that it allows you to do (without attaching dplyr)

```r
library(plotly)
plot_ly(txhousing) %>% 
  group_by_("city") %>% 
  add_lines(x = ~date, y = ~median)
```

I doubt a lot of people are using this, but since this could break user code, we should keep the re-exports

R/utils.R
julianstanley: Addressed this with 4d363d665e30e9e27fb0b2b2dc09d1eb89ebb1e7. 

Big thing: in the rlang issue, lionel recommended using compat-lazy ([see source code here](https://github.com/r-lib/rlang/blob/master/R/compat-lazyeval.R)). It's not in rlang, but an additional ~50 lines that we'd put in utils. It's more generalizable, but more code. 

Should I put compat-lazy into utils, or is dots_as_quosures sufficient?

R/plotly_build.R
julianstanley: Resolved by https://github.com/r-lib/rlang/blob/master/R/compat-lazyeval.R

R/plotly_data.R
julianstanley: Resolved by 4d363d665e30e9e27fb0b2b2dc09d1eb89ebb1e7

R/utils.R
```diff
@@ -1154,3 +1154,21 @@ longest_element <- function(x) {
...
     ""
 }
+
+# Apply rlang::as_quosure across a list or vector, but explicitly pass env
+# and expr to as_quosure
+dots_as_quosures <- function(x) {
+  if (!inherits(x, "lazy_dots")) {
+    stop("Expected lazy dots")
+  }
+  lapply(x, function(x) rlang::as_quosure(x$expr, x$env))
```
cpsievert, line 1164: As is done here https://github.com/r-lib/rlang/blob/64df8e3f/R/compat-lazyeval.R#L51

```suggestion
  lapply(x, function(x) rlang::new_quosure(x$expr, x$env))
```

R/add.R
```diff
@@ -334,7 +334,7 @@ add_sf <- function(p, ..., x = ~x, y = ~y, data = NULL, inherit = TRUE) {
       y = y,
       `_bbox` = bbox,
       set = set,
-      data = if ("group" %in% names(d[[i]])) group_by_(d[[i]], "group", add = TRUE) else d[[i]], 
+      data = if ("group" %in% names(d[[i]])) group_by_add(d[[i]], !!rlang::sym("group"), add = TRUE) else d[[i]],
```
cpsievert, line 337:
```suggestion
      data = if ("group" %in% names(d[[i]])) group_by_add(d[[i]], !!rlang::sym("group")) else d[[i]], 
```

R/plotly_build.R
```diff
@@ -561,7 +561,7 @@ train_data <- function(data, trace) {
     idx2 <- seq.int(2, NROW(dat), by = 2)
     dat[idx2, "x"] <- data[["xend"]]
     dat[idx2, "y"] <- data[["yend"]]
-    data <- dplyr::group_by_(dat, ".plotlyGroupIndex", add = TRUE)
+    data <- group_by_add(dat, !!rlang::sym(".plotlyGroupIndex"), add = TRUE)
```
cpsievert, line 564:
```suggestion
    data <- group_by_add(dat, !!rlang::sym(".plotlyGroupIndex"))
```

cpsievert:
It'd also be great to have a few unit tests to make sure these cases are covered, something like:

```r
expect_warning(
  plot_ly(txhousing) %>% 
    group_by_("city") %>% 
    add_lines(x = ~date, y = ~median),
  NA
)
```

R/add.R
julianstanley: Resolved by fa7768ac7d337b76ad8736ccc1c617e2e389d90e

R/plotly_build.R
julianstanley: Resolved by fa7768ac7d337b76ad8736ccc1c617e2e389d90e

julianstanley:
Weirdly, I am still getting a warning. Thinking it might have been something weird with my system, I switched to a blank RStudio Cloud instance.

I clone my fork, then switch to the patch-1 branch, and install all dependencies. Then:

```R
> devtools::load_all(".")
Loading plotly
Visual testing is not enabled.
> options(lifecycle_verbosity = "warning") 
> txhousing %>% plot_ly(x = ~date, y = ~median) %>% group_by(city)
No trace type specified:
  Based on info supplied, a 'scatter' trace seems appropriate.
  Read more about this trace type -> https://plot.ly/r/reference/#scatter
No scatter mode specifed:
  Setting the mode to markers
  Read more about this attribute -> https://plot.ly/r/reference/#scatter-mode
Warning messages:
1: `group_by_()` is deprecated as of dplyr 0.7.0.
Please use `group_by()` instead.
See vignette('programming') for more help 
2: Ignoring 616 observations 
```

In the terminal, I don't see any instances of `group_by_(` in the R directory:

```bash
rstudio-user@application-2493362-deployment-6489105-2hz46:/cloud/project/plotly/R$ grep -r "group_by_("
plotly_data.R:    #dat <- dplyr::group_by_(dat, crosstalk_key(), add = TRUE)
rstudio-user@application-2493362-deployment-6489105-2hz46:/cloud/project/plotly/R$
```

There are a few in tests, but I don't think those should interfere:

```bash
rstudio-user@application-2493362-deployment-6489105-2hz46:/cloud/project/plotly/tests$ grep -r "group_by_("
testthat/test-plotly-group.R:  dplyr::group_by_("rowname") %>%
testthat/test-ggplot-errorbar.R:    dplyr::group_by_(mtcars, "cyl"),
testthat/test-plotly-data.R:  d <- plotly_data(group_by_(plot_ly(mtcars), c("vs", "am")))
```

Glad that you suggested writing some tests since they caught this. Do you know what might be happening here?

julianstanley:
Oh cool, here's a much more minimal and illustrative example:

```R
> txhousing %>% group_by(city) %>% print(n=3)
# A tibble: 8,602 x 9
# Groups:   city [46]
  city     year month sales  volume median listings inventory  date
  <chr>   <int> <int> <dbl>   <dbl>  <dbl>    <dbl>     <dbl> <dbl>
1 Abilene  2000     1    72 5380000  71400      701       6.3 2000 
2 Abilene  2000     2    98 6505000  58700      746       6.6 2000.
3 Abilene  2000     3   130 9285000  58100      784       6.8 2000.
# … with 8,599 more rows

> txhousing %>% plot_ly() %>% group_by(city) %>% print(n=3)
Error: (converted from warning) `group_by_()` is deprecated as of dplyr 0.7.0.
Please use `group_by()` instead.
See vignette('programming') for more help

> txhousing %>% plot_ly() %>% plotly_data() %>% group_by(city) %>% print(n=3)
# A tibble: 8,602 x 9
# Groups:   city [46]
  city     year month sales  volume median listings inventory  date
  <chr>   <int> <int> <dbl>   <dbl>  <dbl>    <dbl>     <dbl> <dbl>
1 Abilene  2000     1    72 5380000  71400      701       6.3 2000 
2 Abilene  2000     2    98 6505000  58700      746       6.6 2000.
3 Abilene  2000     3   130 9285000  58100      784       6.8 2000.
# … with 8,599 more rows
```

What happens to a plot_ly object when it's passed to `dplyr::group_by`? How is it different than passing to `plotly_data()` and then to `dplyr::group_by()`?

Related to dplyr #5083, I think? https://github.com/tidyverse/dplyr/issues/5083

cpsievert:
The difference in `txhousing %>% plot_ly() %>% group_by(city)` and `txhousing %>% plot_ly() %>% plotly_data() %>% group_by(city)` is that the former winds up calling `group_by_.plotly()` (since the input's class is `plotly`).

It looks like we've made the right changes to `group_by_.plotly()` to avoid this warning now and I've just pulled down your changes to confirm that `txhousing %>% plot_ly() %>% group_by(city)` no longer throws a warning. I'm not sure why you're stilling seeing it...

cpsievert:
Ah, I see why `txhousing %>% plot_ly() %>% group_by(city)` still throws a warning. It's because `group_by(plot_ly())` currently winds up calling `group_by_.plotly()`, and by that time, the warning has been thrown in the generic `group_by_()`

```r
> library(dplyr)
> methods(group_by)
[1] group_by.data.frame* group_by.default*   
see '?methods' for accessing help and source code

> getS3method("group_by", "default")
function (.data, ..., add = FALSE, .drop = group_by_drop_default(.data)) 
{
    group_by_(.data, .dots = compat_as_lazy_dots(...), add = add)
}

> group_by_
function (.data, ..., .dots = list(), add = FALSE) 
{
    lazy_deprec("group_by")
    UseMethod("group_by_")
}
```

This means we'll have to define methods for the non-underscore generic functions found in `R/plotly_data.R` (i.e., we should be defining/registering `group_by.plotly()`). Do you feel up to that task @julianstanley?

julianstanley:
Thanks for following-up on this @cpsievert!

I think that makes sense. If I understand correctly, `group_by` (generally, `[dplyr-verb]`) methods are defined on data.frame (`[dplyr-verb].data.frame` ), but default otherwise (`[dplyr-verb].default`).  So Plotly objects trigger `[dplyr-verb].default`, leading to a warning, but they'll trigger `[dplyr-verb].plotly` if we define that. So we just need to define each SE dplyr method for plotly objects (`[dplyr-verb].plotly`).

Sure thing, this sounds simple enough (famous last words?). I'll give this a shot!

cpsievert:
Thanks for jump starting this effort @julianstanley, I'll take it over the finish line in #1821

julianstanley:
Sounds good, thanks @cpsievert

Issue #1799, opened 2020-06-22
plotly + dplyr verbs + crosstalk; not behaving like before


robertmitchellv:
Before I start; big thanks again to everyone that contributes to plotly on both the r and js side--your work is so appreciated 🙌🏻

I recently updated to the github versions of both plotly and crosstalk and some flexdashboards I maintain for work are not behaving in the way they used to. Here's an example of what used to work:

```r
library(dplyr)
library(plotly)
library(crosstalk)

data <- SharedData$new(diamonds, ~clarity)

bscols(
  widths = c(3, NA),
  filter_select(id = "select", label = "Select clarity", sharedData = data, ~ clarity),
  plot_ly(data) %>%
    group_by(clarity, cut) %>%
    summarise(count = n()) %>%
    add_bars(x = ~cut, y = ~count)
)
```

I've figured out that this now works, which didn't before:

```r
library(dplyr)
library(plotly)
library(crosstalk)

clarity_cut <- diamonds %>%
  group_by(clarity, cut) %>%
  summarise(count = n())

data <- SharedData$new(clarity_cut, ~clarity)

bscols(
  widths = c(3, NA),
  filter_select(id = "select", label = "Select clarity", sharedData = data, ~ clarity),
  plot_ly(data) %>%
    add_bars(x = ~cut, y = ~count)
)
```

I'm just curious if the second code chunk I shared is the way things will be set up moving forward or if this change was unintended.

I wanted to check before I begin to refactor code since I didn't see anything indicating this change was intentional.

cpsievert:
I'm a little surprised either of these have ever worked. Once you aggregate the data on the R side, it's too late for the aggregation to happen client-side. I think what you want here is `add_histogram()` instead of `add_bars()`:

```r
library(dplyr)
library(plotly)
library(crosstalk)

data <- SharedData$new(diamonds, ~clarity)

bscols(
  widths = c(3, NA),
  filter_select(id = "select", label = "Select clarity", sharedData = data, ~ clarity),
  plot_ly(data) %>%
    add_histogram(x = ~cut)
)
```

See more examples of "statistical queries" at https://plotly-r.com/client-side-linking.html#statistical-queries

robertmitchellv:
oof.

I didn't realize that they shouldn't be working at all the the way I was using them. I'll read the section you suggest and change `add_bars()` to `add_histogram()`. Thanks for the help!

cpsievert:
[...]

PR #1821, opened 2020-07-21
Dplyr 1.0

### LINKED ISSUE #1783: Warning: `arrange_()` is deprecated as of dplyr 0.7.0. Please use `arrange()` instead.

cpsievert:
Closes #1783, #1825, and #1799


R/ggplotly.R
```diff
@@ -1112,7 +1112,7 @@ gg2list <- function(p, width = NULL, height = NULL,
   # translate group aesthetics to data attributes
   return_dat <- Map(function(x, y) {
     if (is.null(y[["group"]])) return(x)
-    dplyr::group_by_(x, y[["group"]])
+    dplyr::group_by(x, !!rlang::as_quosure(y[["group"]]))
```
cpsievert, line 1115: TODO: should this be `sym()` instead of `as_quosure()`?

cpsievert: Ah, no, make sense because `y` is formula

DESCRIPTION
```diff
@@ -75,7 +75,7 @@ Suggests:
     forcats,
```
schloerke, line 75: Need to add a version requirement for `  dplyr (>= 1.0.0)`

cpsievert: 1.0.0 isn't actually required

schloerke: The comment is for `dplyr`. I can't comment up there

cpsievert: I know, I don't want to require dplyr 1.0, it should work with older versions too

R/onLoad.R
```diff
@@ -0,0 +1,41 @@
...
+  # Always register hook in case pkg is loaded at some
+  # point the future (or, potentially, but less commonly,
+  # unloaded & reloaded)
+  setHook(
+    packageEvent(pkg, "onLoad"),
+    function(...) {
+      registerS3method(generic, class, fun, envir = asNamespace(pkg))
+    }
+  )
+}
```
schloerke, line 41: Consider copying from `vctrs`: https://github.com/r-lib/vctrs/blob/df8a659135149fb7c97475c80b4f0e3703a5d784/R/register-s3.R#L49-L105

Docs: https://vctrs.r-lib.org/reference/s3_register.html

Via: https://github.com/r-lib/roxygen2/issues/623

R/plotly_data.R
```diff
@@ -104,116 +104,161 @@ highlight_key <- function(...) {
...
 #' @rdname plotly_data
-#' @export
 groups.plotly <- function(x) {
-  dplyr::groups(plotly_data(x))
+  groups(plotly_data(x))
 }
 
 #' @rdname plotly_data
-#' @export
 ungroup.plotly <- function(x, ...) {
```
schloerke, line 117: This is now unexported?  I'm fine with it as long as it still works

R/utils.R
```diff
@@ -139,7 +139,7 @@ crosstalk_key <- function() ".crossTalkKey"
 # arrange data if the vars exist, don't throw error if they don't
 arrange_safe <- function(data, vars) {
   vars <- vars[vars %in% names(data)]
-  if (length(vars)) dplyr::arrange_(data, .dots = vars) else data
+  if (length(vars)) dplyr::arrange(data, !!!rlang::syms(vars)) else data
```
schloerke, line 142: Does `rlang` automatically import `!!!`? Does it need to be imported/accessed manually?

R/plotly_data.R
cpsievert: Yea, the method is now registered in `.onLoad()` instead

R/utils.R
cpsievert: I did end up importing, but I don't think it's necessary:

```r
> rlang::`!!`
function (x) 
{
    abort("`!!` can only be used within a quasiquoted argument")
}
```

PR #1848, opened 2020-09-08
Fixed typo - calculcations of shared Y axis was using X axes

okhoma:


cpsievert:
Thanks, do you have minimal example(s) that demonstrate the fix (better yet, can you add some unit test(s))?

okhoma:
@cpsievert, Is it really just a copy/paste error. It's only effect, as far as I can tell, is a warning:

```r
p <- lapply(c("A", "B", "C"), function(x) plot_ly(x = x, y = 1, type = "bar"))
subplot(subplot(p[[1]], p[[2]], shareY = TRUE), p[[3]], shareY = TRUE)
```

```
Warning message:
In seq_len(nrows * unique(xAxisN)) :
  first element used of 'length.out' argument
```

PR #1927, opened 2021-03-07
remove frameOrder warning

FrankBornais:
*fix #1919

PR #1932, opened 2021-03-13
fix traceNames with frames

### LINKED ISSUE #1618: Subscript out of bounds error when trying to combine segment and markers with frame
I am attempting to make an animated dumbbell chart in `plotly` with `R`.

Following the example [here](https://plot.ly/r/dumbbell-plots/) and using the info [here](https://plot.ly/r/animations/), I came up with the following code:

```r
    df <- data.frame(
      category=c('a', 'b', 'c', 'd', 'a', 'b', 'c', 'd'),
      year=c(2000, 2000, 2000, 2000, 2001, 2001, 2001, 2001),
      val_a=c(1,2,2,1,2,5,6,8),
      val_b=c(3,5,4,7,1,9,2,12)
    )
    
    plot_ly(data = df, frame = ~year) %>%
      add_markers(x = ~val_a, y = ~category, name = "Val_A", color = I("red")) %>%
      add_markers(x = ~val_b, y = ~category, name = "Val_B", color = I("blue")) %>%
      add_segments(x = ~val_a, xend = ~val_b, y = ~category, yend = ~category, showlegend=F) %>%
      layout(
        title = "Val A v Val B",
        xaxis = list(title = "Value"), 
        yaxis = list(title = ""),
        margin = list(l = 65)
    )  
```

Unfortunately this throws the following error:
```
    Error in which(idx)[[1]] : subscript out of bounds
    18.
    registerFrames(p, frameMapping = frameMapping)
    17.
    plotly_build.plotly(instance)
    16.
    instance$preRenderHook(instance)
    15.
    createPayload(x)
    14.
    toJSON(createPayload(x))
    13.
    widget_data(x, id)
    12.
    htmltools::tagList(container(htmltools::tagList(x$prepend, widget_html(name = class(x)[1], package = attr(x, "package"), id = id, style = style, class = paste(class(x)[1], "html-widget"), width = sizeInfo$width, height = sizeInfo$height), x$append)), widget_data(x, id), if (!is.null(sizeInfo$runtime)) { ...
    11.
    toHTML(x, standalone = standalone)
    10.
    as.tags.htmlwidget(x, standalone = TRUE)
    9.
    htmltools::as.tags(x, standalone = TRUE)
    8.
    isTag(ui)
    7.
    rewriteTags(x, function(uiObj) { if (isTag(uiObj) || isTagList(uiObj) || is.character(uiObj)) return(uiObj) else return(tagify(as.tags(uiObj))) ...
    6.
    tagify(x)
    5.
    renderTags(html)
    4.
    save_html(html, file = index_html, background = background, libdir = "lib")
    3.
    html_print(htmltools::as.tags(x, standalone = TRUE), viewer = if (view) viewerFunc)
    2.
    print.htmlwidget(x)
    1.
    (function (x, ...) UseMethod("print"))(x)
```
However, if you run the plot code with **either** the markers **or** the segments, it works fine. Why then can I not seem to combine them?

### LINKED ISSUE #1903: Error subscript out of bounds: registerFrames with simple and cumulative animation on same plot
When producing a plot with both cumulative and non-cumulative animations on the same plot where I assign names to both traces, I get the error 
```Error in which(idx)[[1]] : subscript out of bounds```
when trying to print the plot. I've debugged the issue and I think I isolated the problem, although I am not sure on how to fix it.

I believe the problem is here: https://github.com/ropensci/plotly/blob/9bc585ac0d689ad22e30e086ece6073081a9c37b/R/plotly_build.R#L484

I think ```traceNamesMissing``` is supposed to hold the names of the traces that are missing on the particular frame ```d```, but present on other frames. However, when debugging, ```frameTraceNames``` is a list of unique trace names but
```sapply(d, "[[", "name")```
produces a *list* of character vectors in my situation, and therefore ```traceNamesMissing``` gets assigned with the name of *all* the traces of the plot. For my particular situation, I've been able to resolve the issue by replacing line 484 with
```r
 traceNamesMissing <- setdiff(frameTraceNames, unlist(sapply(d, "[[", "name")))
```
However, since I am not that familiar with the internal workings of the package I don't think this solution can work for all cases.

Here is my reprex.
```r
suppressMessages(library(tidyverse))
suppressMessages(library(plotly))
df1<- tibble(frame = seq(1:10)) %>% mutate(x = row_number(), y = 0)
df2<- tibble(rowid = seq(1:10)) %>% mutate(x = row_number(), y = 1) %>%
  split(.$rowid) %>%
  accumulate(function(old, new){
    bind_rows(old, new)
  }) %>%
  bind_rows(.id = "frame") %>%
  mutate(frame = as.numeric(frame))
  


p_noNames<- plot_ly(df1, type = "scatter") %>%
  add_trace(mode = "markers", x = ~x, y = ~y, frame = ~frame) %>%
  add_trace(data = df2, mode = "lines", x = ~x, y = ~y, frame = ~frame)
out_noNames<- plotly_build(p_noNames)
#> No scatter mode specifed:
#>   Setting the mode to markers
#>   Read more about this attribute -> https://plot.ly/r/reference/#scatter-mode

p_withNames<-  plot_ly(df1, type = "scatter") %>%
  add_trace(mode = "markers", x = ~x, y = ~y, frame = ~frame, name= "A") %>%
  add_trace(data = df2, mode = "lines", x = ~x, y = ~y, frame = ~frame, name = "B")
out_withNames<- plotly_build(p_withNames)
#> No scatter mode specifed:
#>   Setting the mode to markers
#>   Read more about this attribute -> https://plot.ly/r/reference/#scatter-mode
#> Error in which(idx)[[1]]: subscript out of bounds
```

FrankBornais:
- Fix #1618 
- Fix #1903


cpsievert:
Thank you! 

If you don't mind, a NEWS item would also be helpful.

cpsievert:
Hmm, do you know what's going on with these differences?

Screen Shot 2021-03-17 at 6 43 07 PM

cpsievert:
Oh I think they might be a result of #1927?

FrankBornais:
> Oh I think they might be a result of #1927?

Yes that's it. Since it kept only 6 symbols... If you want me to undo that just let me know.

cpsievert:
I think most of the changes are fine except for `tests/figs/subplot/subplot-bump-axis-annotation-traces-shared.svg` and `tests/figs/subplot/subplot-bump-axis-annotation-traces.svg`. It looks like you [added a `type = "scatter"` when it really wasn't needed](https://github.com/ropensci/plotly/pull/1927/files#diff-ad5721ead92ebdbc7912ba83b74e3c208d8152c3d931b5b0206f781ad76e9b18R305-R335) (doing so adds an extra empty trace) -- would you mind reverting that change?

BTW, thank you for updating these visual baselines!!

cpsievert:
Thanks!

PR #1971, opened 2021-07-02
Add kaleido() for static image exporting

cpsievert:
Adds `kaleido()` for static image exporting via https://github.com/plotly/Kaleido/, for example:

```r
library(plotly)
scope <- kaleido()
tmp <- tempfile(fileext = ".png")
scope$transform(plot_ly(x = 1:10), tmp)
file.show(tmp)
```

Using `kaleido()` requires that both the kaleido and plotly Python packages are usable via reticulate. Here's a way to do that via miniconda:

```r
install.packages('reticulate')
reticulate::install_miniconda()
reticulate::conda_install('r-reticulate', 'python-kaleido')
reticulate::conda_install('r-reticulate', 'plotly', channel = 'plotly')
reticulate::use_miniconda('r-reticulate')
```


caozilongsupper:
Hi, I still donnot know how to use it

PR #1999, opened 2021-09-14
Prevent RColorBrewer::brewer.pal warning

s-fleck:
`RColorBrewer::brewer.pal()` throws an (in this case) unnecessary warning if the output plot requires less than 2 colors. This is an easy fix, and does not change the workings of the code since `RColorBrewer::brewer.pal()` never returns less than three levels anywas:

```
> RColorBrewer::brewer.pal(2, "Set2")
[1] "#66C2A5" "#FC8D62" "#8DA0CB"
Warning message:
In RColorBrewer::brewer.pal(2, "Set2") :
  minimal value for n is 3, returning requested palette with 3 different levels
```


cpsievert:
Thanks!

PR #2045, opened 2021-10-13
Added fable/feast support

moutikabdessabour:
Implemented `to_basic` methods that handle the output of `fabletools:autoplot.fbl_ts` and Added tests
```r
library(dplyr)
library(forecast)
library(tsibble)

data <- tourism %>%
    filter(Region == "Melbourne") %>%
    `[`(, c("Quarter", "Trips", "Region")) %>%
    distinct(Quarter, .keep_all = TRUE) %>%
    as_tsibble(key = Region) 
p <- data %>%
    model(
        ets = ETS(Trips ~ trend("A")),
    ) %>%
    forecast(h = "5 years") %>%
    autoplot(data)

p

ggplotly(p)
```

ggplot

ggplotly
.


R/layers2traces.R
```diff
@@ -613,6 +613,41 @@ to_basic.GeomQuantile <- function(data, prestats_data, layout, params, p, ...){
...
+}
+
+to_basic.GeomHiloRibbon <- function(data, ...){
+  prefix_class(transform_hiloribbon(data), "GeomPolygon")
+}
+
+#' @export
+to_basic.data.frame <- function(data, prestats_data, layout, params, p, ...) {
+  prefix_class(data, "GeomPath")
+}
```
cpsievert, line 650: Hmm, this doesn't seem right, do you have an example where this becomes relevant?

R/layers2traces.R
```diff
@@ -613,6 +613,41 @@ to_basic.GeomQuantile <- function(data, prestats_data, layout, params, p, ...){
...
+
+  row_number <- nrow(data)
+
+  data_rev <- data[row_number:1L, ]
+  structure(rbind(
+    cbind(x = data$x, y = data$ymin, data[unused_aes]),
+    cbind(x = data$x[row_number], y = data$ymin[row_number], data[row_number, unused_aes]),
+    cbind(x = data_rev$x, y = data_rev$ymax, data_rev[unused_aes])
+  ), class = class(data))
+}
```
cpsievert, line 641: Come to think of it, I think most of this as well as `to_basic.GeomAlluvium()` could be replaced by `ribbon_dat()`?

cpsievert:
@danton267 would you mind filing an issue to request this feature so that it should up in this list of issues?  https://github.com/plotly/plotly.R/issues/created_by/danton267

R/layers2traces.R
moutikabdessabour: It's the same code with minor changes.

R/layers2traces.R
moutikabdessabour: this is actually a trick because the `autoplot` returns only a data.frame and `GeomHilo...`

cpsievert:
@moutikabdessabour what package is the `model()` function from in your example?

cpsievert:
@moutikabdessabour if we're going to add support for `{fable}`/`{feasts}`, we should try our best to cover anything that one could create via `forecast::forecast()` + `autoplot()`, which would at least include `forecast::GeomForecast`

moutikabdessabour:
@cpsievert Ok I will work on it.

moutikabdessabour:
@cpsievert The `model` function is exported from the `{fabletools}` package

PR #2047, opened 2021-10-14
Adding support to `plotROC`

moutikabdessabour:
- [x] Implement test
- [ ] add changes to `NEWS.md`


R/layers2traces.R
```diff
@@ -622,6 +622,67 @@ to_basic.GeomQuantile <- function(data, prestats_data, layout, params, p, ...){
...
+  text$size <- size
+  text$alpha <- alpha
+  text
+}
+
+#' @export 
+to_basic.GeomRoc <- function(data, prestats_data, layout, params, p, ...){
+  
+  data$alpha <- params$linealpha
+  if(params$n.cuts) {
```
cpsievert, line 639: If this value is falsy, then `points` won't exist, which will lead to an error on line 660, so why have the `if` statement in the first place?

R/layers2traces.R
```diff
@@ -622,6 +622,67 @@ to_basic.GeomQuantile <- function(data, prestats_data, layout, params, p, ...){
...
+    points <- transform_geomroc(
+        data,
+        n.cuts = params$n.cuts,
+        # multiply by 4 to have results similar 
+        size = params$labelsize, 
+        alpha = params$pointalpha, 
+        round = params$labelround
+      )
+    if(params$labels) {
```
cpsievert, line 648: Same comment as above

cpsievert:
The `{plotROC}` already has it's own methods for making interactive plots, so I don't view this as a worthwhile effort https://sachsmc.github.io/plotROC/

PR #2052, opened 2021-10-22
Add stat_ecdf support

moutikabdessabour:
- [x] change `NEWS.md`


R/layers2traces.R
```diff
@@ -623,7 +623,8 @@ to_basic.GeomQuantile <- function(data, prestats_data, layout, params, p, ...){
 }
 
 #' @export
-to_basic.default <- function(data, prestats_data, layout, params, p, ...) {
+to_basic.default<- function(data, prestats_data, layout, params, p, ...) {
+
```
cpsievert, line 627:
```suggestion
to_basic.default <- function(data, prestats_data, layout, params, p, ...) {
```

R/layers2traces.R
```diff
@@ -252,7 +252,7 @@ to_basic.GeomLine <- function(data, prestats_data, layout, params, p, ...) {
 
 #' @export
 to_basic.GeomStep <- function(data, prestats_data, layout, params, p, ...) {
-  prefix_class(data, "GeomPath")
+  prefix_class(if((params$direction %||% "vh")  %in% c("vh", "hv", "mid")) ggplot2:::stairstep(data, direction = params$direction) else data, c("GeomPath", "GeomStep"))
```
cpsievert, line 255: Using `:::` isn't allowed by `R CMD check`. Use `ggfun()` instead

cpsievert: For some reason this breaks default `geom_step()` usage. See, for example, the change in `tests/testthat/_snaps/ggplot-step/step-gg-hv.svg`

moutikabdessabour: I think I fixed the test issues. basically just re rendering the scg solved it

cpsievert: No, the tests are still broken. I don't think this is the right fix.

cpsievert:
This isn't the right approach

PR #2098, opened 2022-01-09
fix error on clearing crosstalk selection

casperhart:
When linking plots using {crosstalk}, running `selectionHandle.clear()` from another (non-plotly) visual causes the following error to appear in the console and the selection in the plotly visual is not cleared:
image

This is because `selectionHandle.clear()` event passes an `undefined` value, which plotly isn't dealing with properly.

This PR fixes this issue.


cpsievert:
Thanks, can you please provide an example that demonstrates the fix?

casperhart:
@cpsievert to reproduce, run the following in R:

```r
library(crosstalk)
library(plotly)
library(DT)

data(iris)

shared_iris <- SharedData$new(iris)

a <- plot_ly(shared_iris, x = ~Sepal.Length, y = ~Sepal.Width) %>%
    add_markers() %>%
    highlight(on = "plotly_click", dynamic = TRUE)
b <- datatable(shared_iris)

bscols(
    list(
        filter_checkbox("species", "Species", shared_iris, ~Species)
    ),
    a, b 
)
```

Then select some points in the plotly visual, open up the browser console and run the following:

```javascript
const handleId = Object.keys(__crosstalk_groups)[1]
const myHandle = new crosstalk.SelectionHandle(handleId)
myHandle.clear()
```

In the main branch code, this throws an error and the selection is not cleared. Using the version in this PR, no error is thrown and the crosstalk selection is successfully cleared.

PR #2131, opened 2022-05-04
Replace `is.na` with `rlang::is_na` in `map_color()`

mojaveazure:
R 4.2 now [throws an error](https://stat.ethz.ch/pipermail/r-announce/2022/000683.html) if the input to `if ()` is has a length greater than 1; this causes errors in `map_color` as one of the conditions in `mapColor` explicitly allows for multiple colors and `is.na` is vectorized. Replacing this with `rlang::is_na` forces the check to essentially be `if (is.na(col) && length(col) == 1)`, thus no longer failing on R 4.2


cpsievert:
Thank you! It'd also be great if you would add a `NEWS.md` item about this!

mojaveazure:
Of course, apologies for not doing that earlier

cpsievert:
Thanks!

Issue #1849, opened 2020-09-12
`ggplotly` modifies values before creating tooltip


dmurdoch:
When drawing a `geom_bar` plot with negative values, a tooltip will display them as positive instead.

```r
library(ggplot2)
library(plotly)
dataset <- data.frame(x = 1:2, y = 1:2 - 3, z = 1:2 - 3) # z == y
p1 <- ggplot(dataset, aes(x, y, z = z)) + geom_bar(stat = "identity")
ggplotly(p1, tooltip = c("y", "z"))
```
Even though `y` and `z` contain the same negative values, only `z` is displayed as negative in the tooltip.

This may be related to #560, #616 or #874.  It was reported on StackOverflow here:  https://stackoverflow.com/q/63862157/2554330

wcwr:
The issue is still unresolved, but adding `geom_point` will at least show the correct value in tooltip :

```
library(ggplot2)
library(plotly)
dataset <- data.frame(x = 1:2, y = 1:2 - 3, z = 1:2 - 3) # z == y
p1 <- ggplot(dataset, aes(x, y, z = z)) + geom_bar(stat = "identity") +
  geom_point(size=4) #Add geom_point
ggplotly(p1, tooltip = c("y", "z"))
```

![image](https://user-images.githubusercontent.com/80007059/199990150-294f8826-9d28-4940-a461-1e4bf598ebaa.png)

Issue #874, opened 2017-02-07
ggplotly does not support negative values


mubashirqasim:
Hi there,

I have just found that ggplotly take absolution of values before plotting. In other words it seems it does not support negative values.

p <- ggplot(dft, aes(x=id, y=value, fill=value)) + geom_bar(stat="identity") +
  geom_abline(intercept = x, slope = 0, color = "blue", size = .3) +
  geom_text(aes(15,x,label = "Avg. value", vjust = -1))

plot(p)
ggplotly(p)

tjbutler:
I believe I found the same thing. I have a histogram with different values below and above the horizontal axis. This plots correctly without plotly, with one histogram downward from the horizontal axis and one above it, but with ggplotly it overlays the two histograms on top of each other, both above the axis.

`set.seed(15)
dd<-data.frame(x=rnorm(100,5,2))
set.seed(5)
dd2<-data.frame(x=rnorm(100,5,2))
p3 <- ggplot() + geom_histogram(data=dd,aes(x=x,y=..count..)) + geom_histogram(data=dd2,aes(x=x,y=-..count..),color='green')
p3
ggplotly(p3)
`
![orig_histogram](https://cloud.githubusercontent.com/assets/2529113/23771460/0f480008-04e5-11e7-82d0-aed9c7fb0f61.png)
![plotly_histogram](https://cloud.githubusercontent.com/assets/2529113/23771613/a477e5da-04e5-11e7-99c5-2c693bdbe147.png)

maxheld83:
I, too, have this issue. I think it is a duplicate of https://github.com/ropensci/plotly/issues/560

rosseji:
Is this not closed because the fix hasn't worked properly? Still getting errors in `plotly_4.7.1`

![screen shot 2017-08-25 at 1 23 31 pm](https://user-images.githubusercontent.com/21113591/29698413-a1eb6dd6-8998-11e7-97a6-ec0a56146e74.png)

cpsievert:
This seems to be fixed now

PR #2201, opened 2022-11-04
Update ggplotly.R

wcwr:
Is use of 'abs' the reason tooltip incorrectly displays positive values as negative? Mentioned in #874 (marked as closed, but problem still persists, as described in more recent #1849 )


wcwr:
NOTE: can be solved by swapping `geom_bar `for `geom_col ` (pointed out by user nirgrahamuk in rstudio community [here](https://community.rstudio.com/t/has-anyone-found-a-workaround-for-the-bug-in-ggplotly-where-negative-values-arent-displayed/151962))

PR #2204, opened 2022-11-09
skip test on pre-3.4.0 ggplot2

MichaelChirico:
on earlier versions, the warning isn't issued.

we could branch the test based on the ggplot2 version, but doesn't seem worth it


cpsievert:
Thank you!

PR #2262, opened 2023-05-02
Handle recent changes to ggplot2's `plot_build()` logic

cpsievert:
Closes #2259

This PR add support for recent (internal) changes to ggplot2 (mostly the ones made in https://github.com/tidyverse/ggplot2/pull/4879 and https://github.com/tidyverse/ggplot2/pull/5144).  These changes removed some internal ggplot2 functions `ggplotly()` was using to translate to plotly.js. Fortunately, the relevant logic is still accessible via methods on the plot object, so going forward `ggplotly()` won't be as reliant on internal ggplot2 functions.

PR #2268, opened 2023-05-19
Relocate tidyverse/ggplot2 remote to `Config/Needs/check`

schloerke:
With #2262 , can the remote be removed?


schloerke:
Moved to `Config/Needs/check` so that testing is done with dev ggplot2, but is not required for installation

PR #2319, opened 2023-11-23
Drop dependency on lazyeval

olivroy:
A fix was inspired by the fix here https://github.com/tidyverse/ggplot2/pull/2797/files.

I fixed small docs issues + updated the link to the repo.

Edit: seealso https://github.com/rstudio/ggvis/pull/488/files


olivroy:
Can't replicate most test failures locally.

This fails in CI (both locally and on GHA), but when I try running it manually, it works fine.

```

PR #2335, opened 2024-01-17
Remove footer from README.md

noamross:
I suggest removing this old rOpenSci footer from the package README, as it can confuse users about the current status of this package.  We're very glad to have been a home for this package in its early days and would be happy if you wanted to note that somewhere in the repo or documentation.


cpsievert:
Thanks!

PR #2336, opened 2024-01-31
Create plotly.js bundle map programmatically

salim-b:
Creates the proper `bundleTraceMap` list as internal package data object directly from the upstream JS file when plotly.js is updated.

The data is sourced from the file [`tasks/util/constants.js`](https://github.com/plotly/plotly.js/blob/ac983b349a1cd4332465f4d791630e8219f42e02/tasks/util/constants.js#L52-L163) of the matching plotly.js Git tag via [`yay::gh_text_file()`](https://yay.rpkg.dev/reference/gh_text_file.html), which I wrote. The package is not (yet) available on CRAN and thus must be installed via `remotes::install_gitlab("rpkg.dev/yay")` for the time being. I think this shouldn't be an issue since [`tools/update_plotlyjs.R`](https://github.com/plotly/plotly.R/blob/master/tools/update_plotlyjs.R) is run manually during development and not intended to be run by users.

This commit adds an updated `bundleTraceMap` for the currently used plotly.js v2.11.1 to `R/sysdata.rda`. It also includes plotly.js' [`strict` bundle](https://github.com/plotly/plotly.js/blob/master/dist/README.md#plotlyjs-strict), so now `plotly::partial_bundle(type = "strict")` works as intended.


DESCRIPTION
```diff
@@ -23,7 +23,7 @@ Description: Create interactive web graphics from 'ggplot2' graphs and/or a cust
 URL: https://plotly-r.com, https://github.com/plotly/plotly.R, https://plotly.com/r/
 BugReports: https://github.com/plotly/plotly.R/issues
 Depends:
-    R (>= 3.2.0),
+    R (>= 3.5),
```
cpsievert, line 26: Is this necessary?

tools/update_plotlyjs.R
```diff
@@ -65,10 +65,28 @@ withr::with_dir(tmpdir, {
...
+      # update plot schema and (partial) bundles
+      if (!nzchar(system.file(package = "yay"))) {
+        stop("pkg 'yay' must be installed via: remotes::install_gitlab('rpkg.dev/yay')")
+      }
       Schema <- jsonlite::fromJSON(Sys.glob("dist/plot-schema.json"))
+      bundleTraceMap <-
+        yay::gh_text_file(path = "tasks/util/constants.js",
+                          owner = "plotly",
+                          name = "plotly.js",
+                          rev = basename(zip)) |>
```
cpsievert, line 77: I think I'd prefer `{gh}` for this

salim-b: `yay::gh_text_file()` uses [`gh::gh_gql()`](https://gh.r-lib.org/reference/gh_gql.html) under the hood. Unfortunately, the maintainers are not up for inclusion of higher-level GraphQL functions, [I've asked](https://github.com/r-lib/gh/issues/173).

salim-b: Addendum: I've just realized that it's easy to fetch a file from a certain Git tag directly via GitHub URLs. For v2.11.1, the URL is:

https://raw.githubusercontent.com/plotly/plotly.js/v2.11.1/tasks/util/constants.js

So, no need for my yay package here. I'll update the PR.

salim-b: I've updated the PR [accordingly](https://github.com/plotly/plotly.R/pull/2336/commits/eaf3aea991b456bfbb5c820e023e56a26921e6bc) and also [added package namespace identifiers](https://github.com/plotly/plotly.R/pull/2336/commits/21c2f0e6a1659b54d84bc4b7a632fb8106edfee2) to make non-base-R dependencies more visible.

salim-b:
@cpsievert Anything else I need to do to get this merged?

DESCRIPTION
cpsievert: The line you're referring to is a build time, not run time dependency, so the version bump requirement isn't necessary

cpsievert, line 26:
```suggestion
    R (>= 3.2.0),
```

tools/update_plotlyjs.R
cpsievert: Thank you, and sorry for the radio silence!

tools/update_plotlyjs.R
```diff
@@ -65,10 +62,29 @@ withr::with_dir(tmpdir, {
...
-      # update plot schema
+      # update plot schema and (partial) bundles
       Schema <- jsonlite::fromJSON(Sys.glob("dist/plot-schema.json"))
+      tmp_file <- tempfile(pattern = "plotly_constants_", fileext = ".js")
+      utils::download.file(
+        url = paste0("https://raw.githubusercontent.com/plotly/plotly.js/", basename(zip), "/tasks/util/constants.js"),
+        destfile = tmp_file,
+        quiet = TRUE,
+        mode = "wb"
+      )
```
cpsievert, line 73: Is this download necessary? Seems like that `tasks/util/constants.js` file should be available locally in the downloaded zip file?

tools/update_plotlyjs.R
```diff
@@ -65,10 +62,29 @@ withr::with_dir(tmpdir, {
...
+      bundleTraceMap <-
+        paste0(readLines(tmp_file), collapse = "\n") |>
+        stringr::str_extract(pattern = "(?<=var partialBundleTraces = )\\{[^}]+\\}") |>
+        yaml::read_yaml(text = _)
```
cpsievert, line 77: It'd be nice if this could also handle `const` as well as throw informative errors if the `str_extract()` comes up empty

DESCRIPTION
salim-b: Yeah, but it also becomes a *run*time dependency once we ship pkg-internal data in v3 format. To cite the [relevant doc](https://usethis.r-lib.org/reference/use_data.html#arg-version):

> Version 3 became the default from R 3.6.0 and **can only be read by R versions 3.5.0 and higher**.

cpsievert: Ahh, right, thanks for pointing that out. Could we just keep the current behavior? It doesn't seem worth it to me to break compatibility for relatively small R objects

salim-b: Keeping the current behaviour is fine with me. But OTOH, R 3.5.0 was [released more than 6 years ago](https://stat.ethz.ch/pipermail/r-announce/2018/000628.html), so I don't think we ought to care about supporting more ancient versions.

I'll have a look at your other comments tomorrow. I'll revert to v2 serialization format then if you still think it's worth it :)

tools/update_plotlyjs.R
salim-b: Addressed in https://github.com/plotly/plotly.R/pull/2336/commits/74a4f4f07159000851a75c68180c0ca313bc511a.

DESCRIPTION
salim-b: Addressed in https://github.com/plotly/plotly.R/pull/2336/commits/93ef777a672852a2aa48c9f7ed5c4c2f995713b5.

salim-b: I've reverted to the v2 serialization format, see https://github.com/plotly/plotly.R/pull/2336/commits/93ef777a672852a2aa48c9f7ed5c4c2f995713b5.

salim-b: I've just realized, you've switched to the v3 format yourself with the latest release (because `usethis::use_data()` defaults to `version = 3` nowadays). Thus, I've bumped the min R version to 3.5.

PR #2351, opened 2024-05-13
Add implicit Suggests 's2' explicitly

MichaelChirico:
As seen here and a few other places:

https://github.com/plotly/plotly.R/blob/44499f21d5610175d2ca65c202fbe07dac41d6f6/tests/testthat/test-ggplot-sf.R#L4

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -65,7 +65,6 @@ Suggests:
     webshot,
     listviewer,
     dendextend,
-    s2,
     sf,
     png,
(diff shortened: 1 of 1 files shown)


cpsievert:
I would love to freely add suggestions for packages used in testing, but unfortunately CRAN limits the number of suggests, so something like `s2` should go under the `Config/Needs/check` field

cpsievert:
Thanks!

PR #2368, opened 2024-07-19
Compatibility with incoming ggplot2 changes

teunbrand:
Hi Carson and other maintainers,

Apologies for the cold PR!
While building a new feature in ggplot2 we found through reverse dependency checks that {plotly} was negatively affected by the changes we'd like to implement. The new feature makes extensive use of computed values in the `Geom$default_aes` field, contrasted by mostly fixed values in the current CRAN version. Mostly, this PR resolves the relevant defaults before using them in {plotly}. The changes herein should be backwards compatible (locally tested with ggplot2 3.5.1).

In addition, a small change with regards to label setup is sneaked in for good measure.
If you rather have a separate PR for that I'll prepare one.

We don't intend to release a new version of ggplot2 this year, so there isn't much of a hurry to merge this.

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -277,6 +277,11 @@ gg2list <- function(p, width = NULL, height = NULL,
     
     # Compute aesthetics to produce data with generalised variable names
     data <- by_layer(function(l, d) l$compute_aesthetics(d, plot))
+    if (exists("setup_plot_labels", envir = asNamespace("ggplot2"))) {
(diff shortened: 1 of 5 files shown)


cpsievert:
> Apologies for the cold PR!

I _very much_ appreciate it -- well done, thank you!!

teunbrand:
Hi @cpsievert, we're aiming for a ggplot2 release before summer (around May). I'd be swell if there'd be a compatible plotly version out around that time :)

PR #2372, opened 2024-07-25
Exclude .git directory and also do not persist Git credentials

mike-sol:


mike-sol:
Have reviewed the produced artifacts and no .git folder is included in there any more.

cpsievert:
Thanks!

Issue #2103, opened 2022-01-31
ggplotly() fails to preserve faceting


AdrianAntico:
If I create a ggplot with faceting, I can end up with several plots in a grid by the facet variables selected. When I convert that output to plotly via ggplotly(), the faceting is removed and the additional plots are blank.

AdrianAntico:
Also, this page should be showing the plot output with faceting but nothing is there

https://plotly.com/ggplot2/facet_grid/

AdrianAntico:
Also, the suggestion link on the top right part of the page doesn't work either

https://github.com/plotly/plotly.r-docs/edit/master/ggplot2/2016-11-29-facet-grid.Rmd

Mayalaroz:
Hello,

I also have a problem with faceting. From this example, 

`p <-    
 ggplot(economics_long, aes(date, value)) +
  geom_line() +
  facet_wrap(vars(variable), scales = "free_y", nrow = 2, strip.position = "bottom") +
  theme(strip.background = element_blank(), strip.placement = "inside")`

When I convert the object p into plotly (via ggplotly), the strip position is not taken into account. It stays in the top.

kwilson62:
Hi, I also have this problem, if I print the ggplot:
![image](https://github.com/user-attachments/assets/e140dff2-14e0-4c5d-8c6a-db15314a8d0d)
`        p <- p + facet_wrap(
          as.formula(paste("~", paste(facet_var_row, collapse = " + "))),
          strip.position = "right",
          scales = "free_y",
          labeller = labeller(.multi_line = FALSE))`
The strip position works correctly, but in ggplotly when rendered in Shiny:
![image](https://github.com/user-attachments/assets/ad80f780-7fb0-4627-b9c8-6b37239ff4fb)

PR #2373, opened 2024-08-05
Bug fix. Account for multi_line=FALSE when facet rows/cols are missing. Closes #2341.

### LINKED ISSUE #2341: `expression(list())` introduced when using labeller in facet
When I am using labeller in facet_wrap I am getting `expression(list())` showing up with ggplotly. 

```r
mtcars$cyl2 <- factor(mtcars$cyl, labels = c("alpha", "beta", "gamma"))
p <- ggplot(mtcars, aes(wt, mpg)) + geom_point()

ggplotly(p + facet_wrap(. ~cyl)) # See figure 1

ggplotly(p + facet_wrap(. ~cyl, labeller = label_wrap_gen(width = 25, multi_line = FALSE))) #see figure2

````

image

image

### PR DESCRIPTION
**Pull request by trekonom**
This PR proposes a fix to account for `multi_line=FALSE` in facet labeller functions, thereby closing #2341 .

The underlying issue is that `ggplotly()` creates a strip text for facet rows/cols even if these are missing, e.g. in case of `facet_wrap` a rows text is created in https://github.com/plotly/plotly.R/blob/3cf17c00477e6992cde648c6e711d5b464b04439/R/ggplotly.R#L937-L941
although in this case `plot$facets$params`  has no `rows` element and hence `plot$facets$params$rows` is `NULL`.

This does not result in an issue as long as `multi_line=TRUE` (which is the default for the `label_xxx` family of functions) as in this case the `label_xxx` family of functions will return an empty `list` which gets "coerced" to an empty string. However, with `multi_line=FALSE` the returned `list` is wrapped in `expression` and as a result gets coerced to a string `"expression(list())"` giving rise to the issue reported in #2341 .

In case of `facet_wrap` this can be fixed by creating a rows text only if `inherits(plot$facet, "FacetGrid")` is `TRUE`. However, the same issue arises with `facet_grid` when the rows/cols are missing:

``` r
library(plotly, warn=FALSE)
#> Loading required package: ggplot2

g <- ggplot(mtcars, aes(mpg, wt)) +
  geom_point() +
  facet_grid(vs + am ~ .,
    labeller = function(x) label_both(x, multi_line = FALSE)
  )

ggplotly()
```

<sup>Created on 2024-08-05 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

To account for this case the PR adds an `if` to check that there is anything to label.

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -917,11 +917,15 @@ gg2list <- function(p, width = NULL, height = NULL,
     # facet strips -> plotly annotations
     if (has_facet(plot)) {
       col_vars <- ifelse(inherits(plot$facet, "FacetWrap"), "facets", "cols")
-      col_txt <- paste(
-        plot$facet$params$labeller(
-          lay[names(plot$facet$params[[col_vars]])]
-        ), collapse = br()
-      )
+      col_txt <- if (!length(names(plot$facet$params[[col_vars]])) == 0) {
+        paste(
+          plot$facet$params$labeller(
+            lay[names(plot$facet$params[[col_vars]])]
+          ), collapse = br()
+        )
+      } else {
+        ""
+      }
       if (is_blank(theme[["strip.text.x"]])) col_txt <- ""
       if (inherits(plot$facet, "FacetGrid") && lay$ROW != 1) col_txt <- ""
       if (robust_nchar(col_txt) > 0) {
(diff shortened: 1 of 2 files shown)


trekonom:
And here is a more detailed description of the issue and the proposed fix:

The underlying issue is that `ggplotly()` creates a strip text for facet rows/cols even if these are missing, e.g. in case of `facet_wrap` a rows text is created in https://github.com/plotly/plotly.R/blob/3cf17c00477e6992cde648c6e711d5b464b04439/R/ggplotly.R#L937-L941
although in this case `plot$facets$params`  has no `rows` element and hence `plot$facets$params$rows` is `NULL`.

This does not result in an issue as long as `multi_line=TRUE` (which is the default for the `label_xxx` family of functions) as in this case the `label_xxx` family of functions will return an empty `list` which gets "coerced" to an empty string. However, with `multi_line=FALSE` the returned `list` is wrapped in `expression` and as a result gets coerced to a string `"expression(list())"` giving rise to the issue reported in #2341 .

In case of `facet_wrap` this can be fixed by creating a rows text only if `inherits(plot$facet, "FacetGrid")` is `TRUE`. However, the same issue arises with `facet_grid` when the rows/cols are missing:

``` r
library(plotly, warn=FALSE)
#> Loading required package: ggplot2

g <- ggplot(mtcars, aes(mpg, wt)) +
  geom_point() +
  facet_grid(vs + am ~ .,
    labeller = function(x) label_both(x, multi_line = FALSE)
  )

ggplotly()
```



<sup>Created on 2024-08-05 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

To account for this case the PR adds an `if` to check that there is anything to label.

R/ggplotly.R
```diff
@@ -917,11 +917,15 @@ gg2list <- function(p, width = NULL, height = NULL,
...
-      )
+      col_txt <- if (!length(names(plot$facet$params[[col_vars]])) == 0) {
+        paste(
+          plot$facet$params$labeller(
+            lay[names(plot$facet$params[[col_vars]])]
+          ), collapse = br()
+        )
+      } else {
+        ""
+      }
```
cpsievert, line 928: This style leads to a smaller diff and also matches the style of the subsequent code

```suggestion
      col_txt <- paste(
        plot$facet$params$labeller(
          lay[names(plot$facet$params[[col_vars]])]
        ), collapse = br()
      )
      if (length(names(plot$facet$params[[col_vars]])) == 0) col_txt <- ""
```

R/ggplotly.R
```diff
@@ -934,22 +938,30 @@ gg2list <- function(p, width = NULL, height = NULL,
...
+        row_txt <- if (!length(names(plot$facet$params$rows)) == 0) {
+          paste(
+            plot$facet$params$labeller(
+              lay[names(plot$facet$params$rows)]
+            ),
+            collapse = br()
+          )
+        } else {
+          ""
+        }
```
cpsievert, line 952: Similar recommendation as https://github.com/plotly/plotly.R/pull/2373/files#r1742335996

cpsievert:
Thank you for the detailed explanation and contribution! This is looking very close to mergeable, I just had a couple stylistic suggestions

kwilson62:
Looking forward to this getting fixed, currently experiencing the issue with "expression(list())"

kwilson62:
@trekonom In addition to the bug you've fixed, there's also an issue where strip.position is not working in plotly.
Considering this issue has been around for awhile now, as seen in this issue:
[#2103](https://github.com/plotly/plotly.R/issues/2103), and you seem to be the most recent / successful at fixing some of these facet issues, do you think you could take a crack at it?

R/ggplotly.R
trekonom: Thx for the review and the suggestions. Just committed both suggestions.

trekonom:
@kwilson62 Will have a look if I find the time. But first let's finish this PR. (:

PR #2387, opened 2024-08-31
Fix. Account for the "data_array" property of the "values" attribute of the "dimensions" attribute. Closes #2385

### LINKED ISSUE #2385: plotly parcoords (parallel coordinates) cannot plot a single line
The following generates no plot, no warnings and no errors:

```r
# Example from https://plotly.com/r/parallel-coordinates-plot/
library(plotly)
df <- read.csv("https://raw.githubusercontent.com/bcdunbar/datasets/master/iris.csv")
df <- df[1L, , drop=FALSE] # plot one line only
fig <- df %>% plot_ly(type = 'parcoords',
          line = list(color = ~species_id,
                      colorscale = list(c(0,'red'),c(0.5,'green'),c(1,'blue'))),
          dimensions = list(
            list(range = c(2,4.5),
                 label = 'Sepal Width', values = ~sepal_width),
            list(range = c(4,8),
                 constraintrange = c(5,6),
                 label = 'Sepal Length', values = ~sepal_length),
            list(range = c(0,2.5),
                 label = 'Petal Width', values = ~petal_width),
            list(range = c(1,7),
                 label = 'Petal Length', values = ~petal_length)
            )
          )
fig
```

trekonom:
This PR proposes a fix to account for the "data_array" property of the "values" attribute of the "dimensions" attribute used in special traces like  "parcoords", "parcats" and "splom" traces thereby closing #2385.

Currently, when an attribute has the "data_array" attribute it gets wrapped in `AsIs` inside `verify_attr`. However, this check gets skipped for the `dimensions` attribute as it is an unnamed list. Additionally, the schema specs for the `values` attribute are somewhat nested.

To account for that the PR adds some special treatment for `dimensions` attribute to `verify_attr` which uses `lapply` to loop over the items of the `dimensions` list.

With the proposed fix the examples documented in #2385 work fine, e.g. for `"parcats"`:

``` r
library(plotly, warn=FALSE)
#> Loading required package: ggplot2

dd <- data.frame(
  from = "A",
  to = "B"
)

plot_ly(dd) %>%
  add_trace(
    type = "parcats",
    dimensions = list(
      list(values = ~from),
      list(values = ~to)
    )
  )
```

<sup>Created on 2024-08-31 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,8 @@
 
 * Closed #2337: Creating a new `event_data()` handler no longer causes a spurious reactive update of existing `event_data()`s. (#2339)
 
+* Closed #2385: `"parcoords"`, `"parcats"` and `"splom"` traces now work with one-dimensional data by accounting for the "data_array" property of the "values" attribute of "dimensions".
+
 # 4.10.4
 
 ## Improvements
--- a/R/utils.R
+++ b/R/utils.R
@@ -551,7 +551,15 @@ verify_attr <- function(proposed, schema, layoutAttr = FALSE) {
     
     # do the same for "sub-attributes"
     if (identical(role, "object") && is.recursive(proposed[[attr]])) {
-      proposed[[attr]] <- verify_attr(proposed[[attr]], attrSchema, layoutAttr = layoutAttr)
+      # The "dimensions" attribute requires a special treatment as 
+      # it is an unnamed list and hence will be skipped by `for (attr in names(proposed))
+      if (attr == "dimensions") {
+        proposed[[attr]] <- lapply(proposed[[attr]], \(x) {
+          verify_attr(x, attrSchema$items$dimension, layoutAttr = layoutAttr)
+        })  
+      } else {
+        proposed[[attr]] <- verify_attr(proposed[[attr]], attrSchema, layoutAttr = layoutAttr)  
+      }
     }
   }
   
--- a/tests/testthat/test-plotly-parcats.R
+++ b/tests/testthat/test-plotly-parcats.R
@@ -5,25 +5,13 @@ expect_traces <- function(p, n.traces, name) {
   L
 }
 
-test_that("No cartesian axes are supplied to a splom chart", {
-  
-  p <- plot_ly(
-    type = 'splom',
-    dimensions = list(
-      list(values=c(1,2,3), label="A"),
-      list(values=c(2,5,6), label="B")
-    )
-  )
-  
-})
-
 test_that("values property has a class of AsIs", {
   p <- plot_ly(
-    type = "splom",
     dimensions = list(
-      list(values = 1, label = "A"),
-      list(values = 2, label = "B")
-    )
+      list(values = "A"),
+      list(values = "B")
+    ),
+    type = "parcats"
   )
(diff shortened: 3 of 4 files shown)


R/utils.R
```diff
@@ -551,7 +551,15 @@ verify_attr <- function(proposed, schema, layoutAttr = FALSE) {
     
     # do the same for "sub-attributes"
     if (identical(role, "object") && is.recursive(proposed[[attr]])) {
-      proposed[[attr]] <- verify_attr(proposed[[attr]], attrSchema, layoutAttr = layoutAttr)
+      # The "dimensions" attribute requires a special treatment as 
+      # it is an unnamed list and hence will be skipped by `for (attr in names(proposed))
+      if (attr == "dimensions") {
```
cpsievert, line 556: Seems there is actually a more general problem here that's beyond just `dimensions`. Some attributes say `role: object`, but are actually expecting a list of objects. Here is the schema for both `dimensions` and `transforms`:

Screenshot 2024-09-03 at 5 10 03 PM

Screenshot 2024-09-03 at 5 10 03 PM


So, I think this code should be closer to:

```r
if (identical(role, "object") && is.recursive(proposed[[attr]])) {
      # Some attributes (e.g., dimensions, transforms) are actually 
      # a list of objects (even though they, confusingly, have role: object)
      # In those cases, we actually want to verify each list element
      attr_ <- sub("s$", "", attr)
      is_list_attr <- ("items" %in% names(attrSchema)) && 
        (attr_ %in% names(attrSchema$items))
      if (is_list_attr) {
        proposed[[attr]] <- lapply(proposed[[attr]], function(x) {
          verify_attr(x, attrSchema$items[[attr_]])
        })
      } else {
        proposed[[attr]] <- verify_attr(proposed[[attr]], attrSchema, layoutAttr = layoutAttr)
      }
    }
```

R/utils.R
```diff
@@ -551,7 +551,15 @@ verify_attr <- function(proposed, schema, layoutAttr = FALSE) {
...
+        proposed[[attr]] <- lapply(proposed[[attr]], \(x) {
```
cpsievert, line 557: `\(x)` was added fairly recently to R, so please use `function(x)` instead

NEWS.md
```diff
@@ -12,6 +12,8 @@
 
 * Closed #2337: Creating a new `event_data()` handler no longer causes a spurious reactive update of existing `event_data()`s. (#2339)
 
+* Closed #2385: `"parcoords"`, `"parcats"` and `"splom"` traces now work with one-dimensional data by accounting for the "data_array" property of the "values" attribute of "dimensions".
```
cpsievert, line 15:
```suggestion
* Closed #2385: `"parcoords"`, `"parcats"` and `"splom"` traces now work when `values` is a length 1 vector (also, more generally, attributes that expect a list of objects should have those objects "verified" correctly now).
```

R/utils.R
trekonom: 🙈 Thanks for the reminder. Should have thought of that myself.

R/utils.R
trekonom: Thx for the review and the suggestions. Just added and comitted.

PR #2404, opened 2024-11-30
Support factor variables for x and/or y when numeric variable is used to set color

BeppoBrem:
Support factor variables for x and/or y when numeric variable is used to set color.

See https://stackoverflow.com/questions/45217616/colors-in-plotly-range-not-meaningful-for-factors. The proposed workaround does not honour the order of the factor levels.


BeppoBrem:
@cpsievert Can you advise on next steps to get this PR reviewed and eventually merged? Many thanks!

PR #2440, opened 2025-06-05
Compatibility ggplot2 4.0.0

teunbrand:
Hi Carson,

We're preparing a new major version of ggplot2 and during a revdepcheck we bumped into a few things we think are best adressed in plotly rather than ggplot2.
I'm still getting a few node stack errors that I can't trace to ggplot2 or reproduce when re-running the relevant test file, so perhaps it'll be false alarm.
We plan to release the new ggplot2 version soon, so it'd be great if a compatible version of plotly was already on CRAN.

Best wishes,
Teun

### DIFF
--- a/R/ggplotly.R
+++ b/R/ggplotly.R
@@ -784,7 +784,7 @@ gg2list <- function(p, width = NULL, height = NULL,
       # the logic here is similar to what p$coordinates$aspect() does,
       # but the ratio is scaled to the data range by plotly.js 
       fixed_coords <- c("CoordSf", "CoordFixed", "CoordMap", "CoordQuickmap")
-      is_fixed <- !is.null(p$coordinates$is_free())
+      is_fixed <- isFALSE(p$coordinates$is_free())
       if (inherits(p$coordinates, fixed_coords) || is_fixed) {
         axisObj$scaleanchor <- anchor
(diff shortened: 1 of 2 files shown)


cpsievert:
Hey @teunbrand, thanks for this! Turns out there were a few more issues that the visual tests surfaced -- I'll fix this and those other things in #2440.

teunbrand:
Ah the visual tests didn't run locally on my machine, so I was blind for any changes therein (I apologise!) I'll have to mention that the most invasive change we plan for 4.0.0 is not merged into the main branch yet. That change is [ggplot2/#6364](https://github.com/tidyverse/ggplot2/pull/6364), and we're polishing that to minimise reverse dependency breakage. So if you're testing against dev ggplot2, then it'd be great if you could test against that PR.

teunbrand:
This is probably a side effect of {GGally} that overrides `+.gg`.
If you `skip()` the following test, these errors should no longer occur.

https://github.com/plotly/plotly.R/blob/7a0176de04bce10f3ecb6aeb46c2981671eb7554/tests/testthat/test-plotly-subplot.R#L137

GGally probably needs to adapt to the new version too, but the C stack errors are not plotly's fault.

PR #2444, opened 2025-06-18
Add valid attribute fill for trace type "box"

oliver-gray:
The "fill" attribute can be used in combination with the box trace type to provide colours to box plots. This is particularly useful to enable corroboration between traces of different types.
For example, between colours used for point/scatter traces and box traces. 
This change sets fill as a valid attribute where thisTrace$type == "box". 

Including this attribute as valid in the source .json file may be an alternative fix for the inaccurate/non-specific warning message.


oliver-gray:
@cpsievert @KasperSkytte Please could you take a look at this? Many thanks.

oliver-gray:
@leamkhoueiry FYI

PR #2486, opened 2026-04-10
`collaborate` is no longer a supported option for `config` and should not be included in the examples

wkumler:
Ran into an unexpected warning when trying to run the `config` examples because the `collaborate` argument that it uses has since been deprecated. This PR removes that arg from the example and uses a different one (disabling the "lasso" option via `modeBarButtonsToRemove = list("lasso2d")`) instead. Unit test for the "collaborate" warning has been added and docs have been rebuilt for that part of the file specifically.


cpsievert:
Thanks!

